PR #1, opened 2015-02-05
Setting up front-end base

nhqe:
Setting up the front-end base.

Notes:
Uses Mixture.io, (now free) application to compile & minify Sass amongst other benefits. Using Mixture, open the /site/ directory and 'View Locally' to see the in-progress.

Will also generate the static HTML at a later date.


sstangl:
Mixture only supports Mac and Windows. It's provided gratis, but the source code isn't available, so it's not even free software. Since I can't run this program and requiring non-free software goes against the spirit of this project, we should find something else.

You could create the site using Mixture and then upload the static documents to the repository. This would be acceptable if they still in a form that's easy to modify. But it sounds like this involves minification, and the repository shouldn't include compiled or minified code.

PR #8, opened 2016-06-10
Searching multiple matches

KDercksen:
The search behaviour works as intended, however the UI hint is not going to work properly unless I use jQuery (or I'm missing something). Not sure if that's okay with you? This is what I've got so far.

### DIFF
--- a/web/scripts/app.js
+++ b/web/scripts/app.js
@@ -4,7 +4,7 @@
 var grid; // The SlickGrid.
 var sortCol = {id: 'wilks'}; // Initial column sorting information.
 var sortAsc = false; // Initial column sorting information.
-var searchInfo = {lastrow: 0, laststr: ''};
+var searchInfo = {laststr: ''};
 
 var theTable = document.getElementById("thetable");
 var boxRaw = document.getElementById("raw");
@@ -205,7 +205,6 @@ function redraw() {
 function _search_from(query, rowid) {
     var data = grid.getData();
     var numrows = data.getLength();
-    var next = false;
 
     for (var i = rowid + 1; i < numrows; ++i) {
         var row = data.getItem(i);
@@ -225,12 +224,9 @@ function search() {
     if (!query)
         return;
 
-    var rowid;
-    if (query === searchInfo.laststr) {
-        rowid = searchInfo.lastrow;
-        searchbutton.textContent = "Next";
-    } else {
-        searchbutton.textContent = "Search";
+    var rowid = grid.getViewport().top;
+
+    if (query !== searchInfo.laststr) {
         rowid = -1;
     }
 
@@ -251,7 +247,6 @@ function search() {
 
         // update searchInfo
         searchInfo.laststr = query;
-        searchInfo.lastrow = rowid;
     }
 }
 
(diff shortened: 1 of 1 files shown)


web/scripts/app.js
```diff
@@ -201,6 +202,23 @@ function redraw() {
...
 
 
+function _search_from(query, rowid) {
+    var data = grid.getData();
+    var numrows = data.getLength();
+    var next = false;
+
+    for (var i = rowid + 1; i < numrows; ++i) {
+        var row = data.getItem(i);
+        if (row.searchname == query || row.searchname.indexOf(query) >= 0) {
```
sstangl: These are sort of the same thing. With this new behavior, we probably just want `if (row.searchname.indexOf(query) >= 0)`.

web/scripts/app.js
```diff
@@ -201,6 +202,23 @@ function redraw() {
...
+            rowid = i;
+            next = true;
+            break;
```
sstangl: Since this logic is in a helper function that returns the row index, the above three lines can be replaced with just `return i;`, and the variables `rowid` and `next` may be removed.

web/scripts/app.js
```diff
@@ -4,6 +4,7 @@
 var grid; // The SlickGrid.
 var sortCol = {id: 'wilks'}; // Initial column sorting information.
 var sortAsc = false; // Initial column sorting information.
+var searchInfo = {lastrow: 0, laststr: ''};
```
sstangl: Hm, I actually think this behavior with `lastrow` might be unexpected. Searches in office tools have the concept of "the current point", like a text selection, from which searching proceeds. So if you execute a search, then click some text behind the result and execute the search again, you would expect to get the same result.

This implementation ignores the document state and always proceeds linearly. So if for example someone searches for a name, then scrolls up a little bit and re-executes the search, they would get something else!

Probably this can be fixed by removing the concept of `lastrow` and instead executing the search from the top of the grid plus one row. You can ask the grid for the number of the top row via `grid.getViewport()`, which will return an object with the property `top` set to the top visible row`s ID.

web/scripts/app.js
```diff
@@ -230,7 +230,9 @@ function search() {
     var rowid;
     if (query === searchInfo.laststr) {
         rowid = searchInfo.lastrow;
+        searchbutton.textContent = "Next";
```
sstangl: It would be `searchbutton.innerHTML = 'Next'`. But maybe we don't actually need a UI hint, and can just always leave the text as 'Search'.

KDercksen:
Using document state makes sense, I agree.

How do you feel about using jQuery for the searchbox? Something like:

```javascript
$("#searchfield").on("keypress change paste", function () { ... })
```

sstangl:
jQuery is OK. I was originally intending to write the site without it, but SlickGrid brought it in anyway as a dependency, so if you'd like to write new code using it, or port old code to it, that would be fine. I don't know it myself, but it seems pretty intuitive. Just no jQuery in the code that has to be fast, please!

sstangl:
Just to be clear, I think `laststr` is still needed.

If they do a search, don't change the search text, scroll around, and re-execute the search, we should probably return the "next" result based on the viewport.

But if they do a search, then change the search text, people probably don't expect "from this point" anymore: they are probably expecting a search from the top.

sstangl:
Thanks, looks good!

PR #11, opened 2016-06-13
Allow scrolling with PageUp/Down (fixes #10)

### LINKED ISSUE #10: Page Up doesn't work
Because SlickGrid is non-selectable, it also broke Page Up / Page Down.

KDercksen:
Should speak for itself.

### DIFF
--- a/web/scripts/app.js
+++ b/web/scripts/app.js
@@ -260,6 +260,15 @@ function searchOnEnter(keyevent) {
     }
 }
 
+function scrollOnPageUpDown(keyevent) {
+    console.log(keyevent.key);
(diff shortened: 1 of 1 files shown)


sstangl:
Wow, thank you so much!

sstangl:
I wonder if there's a way to attach a CSS animation to make it scroll smoothly. SlickGrid maintains a buffer of out-of-screen rows to prevent flickering, so theoretically the DOM target should exist.

sstangl:
The current behavior is really great, though. I especially like that you can leave focus on the search box and still scroll around with pageup/pagedown. Thanks for the patches!

PR #35, opened 2016-09-09
Design tweaks

nhqe:


sstangl:
Thank you so much!

I kept all of the changes except for the background color, which I made 0x040404. At least on my monitor, that seems to render less painfully than pure black. You're right that it could be easier on the eyes.

I also got the current page URL highlighted in red (the one you chose) again. Hopefully I did that correctly!

The changes are live on the website as of right now. Thank you again!

PR #64, opened 2017-04-24
Update lifters.csv

emp100k:
Changed equipment from raw to Multi, as this would be an all time world record it is highly peculiar that no one has reported on it.  Also I asked another lifter from this meet through personal message if Jorge  broke a world record at this meet and they said "definitely not".  Likely just a simple notation error.


meet-data/rps/1678/lifters.csv
```diff
@@ -80,8 +80,8 @@ M,Wraps,Amateur Open,125,Elliott Minshew,115.8,217.5,125,217.5,560,SBD
 M,Raw,Amateur Open,140,Lee Curtis,129.3,137.5,105,165,407.5,SBD
 M,Wraps,Amateur Open,140,Colt Hobbs,136.7,275,182.5,277.5,735,SBD
 M,Wraps,Pro Open,140,Vincent Vela-Pulido,135.1,330,212.5,342.5,885,SBD
-M,Raw,Amateur Open,82.5,Jorge Stergiou,80.9,,262.5,,262.5,B
-M,Raw,Amateur Submaster,82.5,Jorge Stergiou,80.9,,262.5,,262.5,B
+M,Multi,Amateur Open,82.5,Jorge Stergiou,80.9,,262.5,,262.5,B
+M,Multi,Amateur Submaster,82.5,Jorge Stergiou,80.9,,262.5,,262.5,B
```
sstangl: The string is required to be "Multi-ply".

PR #65, opened 2017-04-24
Update lifters.csv

emp100k:
Changed Micheal Bingham's bench press result form Raw to Single-Ply.  If his result was raw not only would he have broken a world record but at the age of 49 he would have broken a 30 year old record by 50 lb, clearly this was miss labeled as Raw rather than Single Ply.
Also researching this man's name I could come up with no news of a WR or video of the lift so it's very unlikely that this was a Raw lift.


meet-data/uspa/0427/lifters.csv
```diff
@@ -123,8 +123,8 @@ DQ,Mark Lohner,TX,100,99.34,48,-235,,,,,M,Master 45-49,Single-ply,,,
 1,Tim Metcalf,TX,82.5,79.47,54,,112.5,,112.5,77.13,M,Master 50-54,Raw,,,
 1,Russell Richter,TX,100,97.34,50,,177.5,,177.5,109.24,M,Master 50-54,Raw,,,
 1,Brian Propst,TX,125,110.4,47,,202.5,,202.5,119.04,M,Master 45-49,Raw,,,
-1,Micheal Bingham,PA,82.5,80.92,49,,275,,275,186.404,M,Open,Raw,,,
-1,Micheal Bingham,PA,82.5,80.92,49,,275,,275,186.41,M,Master 45-49,Raw,,,
+1,Micheal Bingham,PA,82.5,80.92,49,,275,,275,186.404,M,Open,Single,,,
+1,Micheal Bingham,PA,82.5,80.92,49,,275,,275,186.41,M,Master 45-49,Single-ply,,,
```
sstangl: Looking at the results.txt, he competed Multi-ply.

PR #66, opened 2017-04-24
Update lifters.csv

emp100k:
Added BodyweightKg column in order to add a body-wight for April Mathis (in order to gain an approximate Wilks score and sort this result properly into the correct weight class.)  Her approximate weight was extrapolated from the average of the meet before (259 lb) and meet after (263 lb) to get a rough idea of her weight for this meet. (261 lb)


sstangl:
No speculative data, unfortunately. If the federation didn't report it, it shouldn't be in the database. The same thing happened to Malanichev at ProRaw BigDogs -- everyone was listed as SHW and no one's bodyweight was recorded. I think he would have set some Wilks record had it been.

emp100k:
Fair enough, I figured that might be the case, thanks anyhow.

PR #72, opened 2017-04-24
Update lifters.csv

emp100k:
Changed David Douglas from Raw to single ply, the results.txt shows him in the single-ply division and his bench numbers would have otherwise been above the world record if his lifts were raw.


sstangl:
Thanks!

PR #73, opened 2017-04-24
Update lifters.csv

emp100k:
Changed J.B Hill from Raw to Single-ply, his bench press would be a 50lb world record in that weight class.  Unfortunately did not know whether to change it to Single-ply or Multi-ply as the results document doesn't specify.


emp100k:
Makes sense, I will change it now.

sstangl:
Thanks :-)

PR #74, opened 2017-04-27
Update lifters.csv

emp100k:
Changed shana miller, jen millican, ellen stein, sarah cattan, jenine whittaker, laura hauffman, Beth Tarkany, amy johnson, Irina Petrovich, Kellyn Huehn, Caitlin Holmstrom and Sabrina Provoast from Raw to Wraps.
Looks like some of the girls in the 2016 SPF Women;s Pro AM are miss-assigned raw when they were in wraps a few of their squats results are inordinately high I don't know for sure how many were miss labeled, but these one were the ones that stuck out most (based on prior performance and compared to current raw without wraps squat records).
Actually a pretty good video here with bio cards that helped me confirm that these lifters were in wraps: https://www.youtube.com/watch?v=WityO2AEpxM


sstangl:
Thanks! I'm going to run through and mark the rest of them Wraps as well. Clear error on my part.

emp100k:
I mean it's easy to do, don't think SPF ever differentiates between raw and wraps in their results sheets.

sstangl:
Yeah, they don't. Only way to catch this kind of error is manually. Thanks again!

PR #77, opened 2017-04-27
Update lifters.csv

emp100k:
Andy's name is listed 3 different ways on the database, substituting the one with the most meet results "Askow Andy" for Andy Askow or Andrew Askow in 3 meets.


sstangl:
Same as #78, should be "Andy Askow" based on his Instagram.

sstangl:
Fixed by other pull request.

PR #78, opened 2017-04-27
Update lifters.csv

emp100k:
Andy's name is listed 3 different ways on the database, substituting the one with the most meet results "Askow Andy" for Andy Askow or Andrew Askow in 3 meets.


emp100k:
Yeah that's the problem, guess I'll have to go change all those meets to Andy Askow, should only be about 6 meets I think...

sstangl:
Fixed by other pull request.

PR #79, opened 2017-04-27
Update lifters.csv

emp100k:
In all other cases Liz Craven's name shows up as Elizabeth Craven so I changed it to that.


sstangl:
These are two different lifters: the PA/NZPF Elizabeth Craven is ~40, but the USAPL Liz Craven is ~17.

emp100k:
I think that was actually an error with age notation, otherwise it would be uncanny for two women to have the same name both be in the -52 class and both have top 5 totals.  Also I watched the live stream for  2016 classic worlds that Elizabeth was at, and the Arnold grand prix, so I can vouch that it is the same person.  The age for her on the Arnold grand prix should be changed to ~40.

sstangl:
Ah, you're right. Tricked by the USAPL on two fronts!

Fixed it in f4cf2da5d230965f80583eea008dbb6bc7888776.

PR #98, opened 2017-05-03
Update lifters.csv

emp100k:
Changed Andrew Pears from raw to single ply, not only would his bench result have been a raw USAPL and IPF record but also all time world record regardless of federation in raw full meet bench.  The records database for both IPF an USAPL don't show this so one can safely assume this was a simple miss labeling of equipment class.


meet-data/usapl/WA-2014-02/lifters.csv
```diff
@@ -37,7 +37,7 @@ DQ,F,Aaliyah Gray,B,R-T1,48,Raw,46.9,,,,
 1,M,Kevin Greene,SBD,R-JR,75,Raw,74.6,170,100,170,440
 1,M,Eric Wuthrich,SBD,R-JR,82.5,Raw,76.7,140,100,180,420
 1,M,Sun Ly,SBD,R-JR,90,Raw,86.4,182.5,135,205,522.5
-1,M,Andrew Pears,SBD,R-JR,100,Raw,95.5,242.5,275,292.5,810
+1,M,Andrew Pears,SBD,R-JR,100,Single-ply,95.5,242.5,275,292.5,810
```
sstangl: Change seems plausible. Then his division should be changed from "R-JR" to "JR", and Pierce Frandle should be marked as 1st place in his division.

PR #99, opened 2017-05-03
Update lifters.csv

emp100k:
In two different places on the score-sheet it shows Derrington Wright's bench press result as 193.5 and then 255 kg.  Either one of these entries is in error or the higher number was a 4th attempt in a bench shirt.  Both were labeled as Raw the only differentiation is one is under Open and the other is under Junior.  It's not uncommon for lifters to be entered into multiple weight classes, but the higher bench press is more than 120 lb higher than any of Derrington's other raw bench results so it appears to be either a equipped attempt or some kind of notation error.  Please advise.


meet-data/uspa/0427/lifters.csv
```diff
@@ -27,7 +27,7 @@ DQ,Leslie Knickman,TX,75,73.39,48,-125,,,,,F,Master 45-49,Raw,,,
 1,Justin Rene Villegas,TX,82.5,81.47,19,192.5,142.5,205,540,364.51,M,Junior 18-19,Raw,212.5,,
 1,Zack Knickman,TX,82.5,82.46,21,215,152.5,275,642.5,430.54,M,Junior 20-23,Raw,,,
 1,Eric Maldonado,TX,90,89.81,18,235,150,280,665,424.99,M,Junior 18-19,Raw,,,
-1,Derrington Wright,TX,90,89.18,22,245,255,295,795,509.91,M,Junior 20-23,Raw,,,
+1,Derrington Wright,TX,90,89.18,22,245,255,295,795,509.91,M,Junior 20-23,Multi-ply,,,
```
sstangl: Luckily for us, Derrington entered two different divisions in this meet. The results for the rows should be the same. The USPA clearly just made an error in this case. Thanks for catching it!

In the Open category, he went 245/193.5/295/733.5.
So I would change the "255" to "193.5", and the "795" to "733.5".

Unfortunately, this then makes the Wilks column incorrect. The easiest way to get the test suite to pass is to delete the entire Wilks column in this file (using a spreadsheet editor).

emp100k: Wouldn't you just copy the Wilks from his other entry?

PR #100, opened 2017-05-03
Update lifters.csv

emp100k:
Changed from 267.5 to 167.5, clearly a notation error on the squat weight, the CPU results website shows the correct number: http://www.powerlifting.ca/nationals/2017/results.php


emp100k:
Oops didn't change the total to reflect the squat, -fixed.

PR #102, opened 2017-05-25
Update name-corrections.dat

emp100k:
Most of these are simple spelling errors, Matt Sohmer goes by Matt not Matthew on his Instagram so I went with that.  As for Shorty, his real name is Bienvenido, but he hasn't used it in a contest registration since 2014 and anywhere you can find him online he goes as Shorty so that name makes most sense.


sstangl:
Thanks!

PR #107, opened 2017-05-30
Update lifters.csv

emp100k:
According to the results data attached to this meet: https://github.com/sstangl/openpowerlifting/blob/master/meet-data/spf-archive/0415/results.csv , Rene Garanta is in the 259 lb (117.5 kg) class, not the 53 kg class.


sstangl:
Thanks!

PR #115, opened 2017-06-14
Create lifters.csv

emp100k:
Looks like there was a little oops on this one, all the varsity single-ply lifters were labeled as raw starting with Trai Chatman and ending with D Hawk. http://usapl.liftingdatabase.com/competitions-view?id=1635


sstangl:
Equipment is inferred from the division (like M-O versus M-OR)... it looks like the USAPL does have the division information in their database, but for whatever reason that particular meet doesn't display it. Thanks for the correction!

PR #117, opened 2017-06-22
Update lifters.csv

emp100k:
Alison lifted multi ply in this meet (was tipped off by an unusually high bench press of 500lb) confirmed it over at her Instagram: https://www.instagram.com/p/BTTi8B6hi7p/


sstangl:
Thanks. SPF made the error here: http://www.southernpowerlifting.com/event/womens-proam-2/

sstangl:
Her Place also had to be updated, did that in another commit.

PR #123, opened 2017-07-18
Update lifters.csv

emp100k:
I think all the weights lifted are in pounds? (not weight classes though)  Otherwise there would be multiple massive world record performances in this meet by young teens 15-17.   do all these weights have to de converted to KG or can they be entered as lbs?


emp100k:
Ok looks like I can't do the columns like that, do you know the best way to fix this meet data without a huge headache?

sstangl:
If you name the columns with "LBS", for example "Bench1LBS", "TotalLBS", etc., you can run the script `script/csv-tokg` over that file and it will convert all the LBS column to Kg columns.

sstangl:
Or I can do it, since that can't be done from the GitHub interface!

sstangl:
Thanks, good catch.

emp100k:
Hey no problem, and thank you with your help fixing it!  :) 

    On Tuesday, July 18, 2017 3:05 PM, Sean Stangl <notifications@github.com> wrote:
 

 Thanks, good catch.—

PR #125, opened 2017-07-19
Update lifters.csv

emp100k:
This push pull meet looks exactly like the other CPF one from earlier today, the numbers for many of the benches and deadlifts exceed all time world records and is highly suspect I'm guessing this is another case of pounds rather than kilos as even in equipment no one has ever deadlifted over 1100 lbs...


sstangl:
Thanks! These high school meets seem to have regularly tricked me. I did like a 14-hour weekend marathon of entering in CPF meets over the weekend, and I probably got burned out by the end.

Hopefully the CPF will link to the site as a result of that work, though.

emp100k:
The thing that confuses me is, why would they list all the weight classes in kilograms but then do the lifts in pounds that doesn't make much sense...  ( about as much sense as a 17 year old deadlifting 1178 lbs...  :P   ) 

    On Wednesday, July 19, 2017 1:32 AM, Sean Stangl <notifications@github.com> wrote:
 

 Thanks! These high school meets seem to have regularly tricked me. I did like a 14-hour weekend marathon of entering in CPF meets over the weekend, and I probably got burned out by the end.Hopefully the CPF will link to the site as a result of that work, though.—

sstangl:
Canada does nearly everything in the Metric system, but they buy their weights from the USA, so most of their gyms are stocked with pound plates. It was pretty surprising to me at the time.

PR #127, opened 2017-07-20
Update lifters.csv

emp100k:
I believe most of the single ply performances here are miss-attributed and are actually raw.   Based on the roster at https://docs.google.com/spreadsheets/d/1_4mRTRG9k7PWhi95sClcIazVllUALoULc6cY--WvOFw/edit#gid=519913814
there were only like 5-6 single-ply lifters.
With this meet I think the scorecard on http://usapl.liftingdatabase.com/competitions-view?id=1700 was organized n a weird way due to the fact that there were some lifters doing a 5-lift super total and others that were only doing powerlifting...


sstangl:
Thanks. Lots of problems with that meet. USAPL's data quality has been going way down -- it looks like they're having meet directors enter the results in themselves instead of having it done by a central location, and they're making errors all over the place.

emp100k:
Well in this case I would probably chalk it up to the very special kind of meet that 5-bar showdown was (combining both powerlifting and olympic lifting) that may have affected the way they did the score-sheet as they had to keep the lifters competing for the overall 5 lift total separate. 

    On Wednesday, July 19, 2017 11:44 PM, Sean Stangl <notifications@github.com> wrote:
 

 Thanks. Lots of problems with that meet. USAPL's data quality has been going way down -- it looks like they're having meet directors enter the results in themselves instead of having it done by a central location, and they're making errors all over the place.—

Issue #16, opened 2016-06-17
Distinguish multiple lifters with the same name


sstangl:
Finally the annoying thing happened: http://www.openpowerlifting.org/lifters.html?q=Emily%20Hu

The "Emily Hu" competing in Ohio and Massachusetts is different from the one competing in the other meets. We need some way to tell these lifters apart. That probably means that lifters need to be given a unique ID, although we don't want to do that across the entire database because then everything gets super unreadable.

Probably the way this should work is as follows:
- Appending `#` to a name specifies a unique ID for lifters of that name. So there would be `Emily Hu #1` and `Emily Hu #2`. The `#` mark is required since some federations append numbers to names to denote multiple entries, and we don't want to accidentally give those numbers significance.
- Lifters who don't share their names with anyone, which should be most people, can just omit the number like we've been doing.
- Tooling can scan every `lifters.csv` and find the list of lifters who require number specifiers. If a name in that list shows up without an ID, it can warn.
- There needs to be a way to specify "I have no idea which one this is", at which point that lifter's data gets added to all the other ones. Perhaps `#?`.
- Maybe some of the tagging can be updated based on age information -- but only USPA provides that.
- Rendering the lifter's name on the webpage just removes the number.

sstangl:
Also now there are two "Jennifer Thompsons", one in the UPA.

sstangl:
See also #119.

PR #128, opened 2017-07-20
Update lifters.csv

emp100k:
Damien Garrity is 22 not 8, (1300+ lb total is not possible for an 8 year old) also his other meet in 2017 he is 23.
https://github.com/sstangl/openpowerlifting/blob/master/meet-data/pa/1734/results.csv


sstangl:
Thanks! I'll file an issue to throw an error on pre-teens with big totals so this doesn't happen again.

emp100k:
In a unrelated issue I noticed at least one name that is the exact same for two different lifters in the same weight class, for instance Clinton Lee is two different men that lift in the -83kg class, one is 23, from New Zealand and has totals in the 1500-1600lb range, the other is 32, in the USAPL and has totals in the 1000-1100 range, in the database they are showing up as the same person when clearly it is two different people.  What do you do in the case of two people with the same name spelling and weight class to differentiate? 

    On Thursday, July 20, 2017 3:49 PM, Sean Stangl <notifications@github.com> wrote:
 

 Thanks! I'll file an issue to throw an error on pre-teens with big totals so this doesn't happen again.—

sstangl:
I don't know yet. This is a big problem that comes up a lot. It's filed as Issue #16.

I /think/ it could work to force one of them to be named `Clinton Lee #1` and the other named `Clinton Lee #2`. The build system would then check to make sure that there are no lifters named "Clinton Lee" -- you'd have to specifically label each "Clinton Lee" with a number from that point on.

The site could hide the `#1` and `#2` while rendering the names... but I'm not yet sure how to make it obvious that there are two of them.

Ideas welcome!

PR #141, opened 2017-07-24
Update lifters.csv

emp100k:
Per Leah Davis's Instagram this was a Raw (without wraps) squat/total : https://www.instagram.com/p/BW5Ozk8ASH1/?taken-by=leahrmt


sstangl:
Awesome, thank you so much!

PR #168, opened 2017-07-30
SPF nationals 2007

Borisdatzar:
@sstangl This was a pretty hard one, and it wouldn't surprise me if there is an error. Please check thoroughly if you can


Borisdatzar:
Ho do you know all those benches are single ply?

sstangl:
In the case of this very old meet, it's clear that they used "Open" to mean "Single-ply". That was the usage of the word back in the old days. The IPF/USAPL still use it like that sometimes.

Borisdatzar:
What about Wayne's B entry? Singl-ply also I think it's the same lift as his BD lift.

How do I enter a 4th lift? New Column called  Deadlift4Kg which is blank for everyone else?

sstangl:
I don't see anyone named "Wayne B", but there is a "Wayne Stover" who competed in both "Open Bench Press Only" and "Open Push/Pull", both of which are Single-ply.

Yes -- a new column named Deadlift4Kg that is blank for everyone else.

Borisdatzar:
I'm still very confused. Why is Donald's bench only single ply? He is listed as Raw not open?

sstangl:
Ah sorry, I was looking at his Squat-only, which is Single-ply. You're right that Donald's Bench-only is Raw.

Borisdatzar:
@sstangl please check one more time. I pushed the changes

sstangl:
It's been manually merged.

PR #169, opened 2017-07-30
SPF Gritmas Classic

TrystanOakley:
@sstangl Here is the SPF Gritmas Classic


sstangl:
Awesome, thank you so much!

Please remember that the most important thing is to not get burned out, and only do things if they seem fun. If we slowly chip away at the SPF archive, we'll get it done.

TrystanOakley:
I'm loving it man. Lots of fun and love helping this grow and being a part.

PR #174, opened 2017-08-01
Update lifters.csv

TrystanOakley:
Changed Gabrielle Tucker's last name to Mamani because she got married


sstangl:
There's a much better way to make this kind of change.

First, edit `lifter-data/name-corrections.dat`. The format of each line is: `correct name,misspelling one,misspelling two,...`

So in this case, you would add the line `Gabrielle Mamani,Gabrielle Tucker`.

Then run `scripts/standardize-names lifter-data/name-corrections.dat` from the project root, and it will make all these changes for you in a single commit.

sstangl:
Made that change here: https://github.com/sstangl/openpowerlifting/commit/880db3b01f7d53fa970f4a304af1f9daa90d7a85

TrystanOakley:
Oh thank you. I'll write that down and do it that way next time.

sstangl:
I added documentation here: https://github.com/sstangl/openpowerlifting/blob/master/docs/lifter-name-corrections.md

TrystanOakley:
Thank you.

PR #182, opened 2017-08-02
Hoosier Powerlifting Classic

TrystanOakley:
@sstangl Please look at the lifters Andrew Ingenito and Dominic Duhnauant as they are on there twice with different numbers and events. Also, Sex was only listed for some lifters so I did my best guess. Here is the link to the meet on the SPF website. http://www.southernpowerlifting.com/event/hoosier-powerlifting-classic/


sstangl:
Looks good to me, thanks! I wonder if that guy's Deadlift-Only entry was meant to be a 4th attempt, but it's fine to leave it as the federation reported it.

TrystanOakley:
I'm not sure because it looks like their deadlifts that are in the total
are larger than the "4th attempt"

PR #192, opened 2017-08-08
Arkansas State Powerlifting And Bench Press

TrystanOakley:
@sstangl


sstangl:
Great, thanks! Good job being careful to find all the women... they didn't really make that one easy.

TrystanOakley:
No they like to make them ambiguous😅

PR #209, opened 2017-08-13
Bar Benders Of The Bay

TrystanOakley:
@sstangl


TrystanOakley:
Why does this one look different from the others @sstangl ? It looks like someone already did this meet but it was kept in the todo folder and I redid it or something?

sstangl:
Yes, that seems to be the case.

This was an old meet that used to be in spf/, and was entered before spf-archive/ existed.

When you ran `./helper`, it should have complained something like "mkdir: cannot create directory '0415': File exists." This shouldn't happen too often, but it's worth being aware of.

TrystanOakley:
Yeah it did that. Okay

sstangl:
Resolving as a duplicate, but I'll remove the results file in a separate commit. Sorry about that.

TrystanOakley:
No problem. Shit happens

PR #212, opened 2017-08-15
Changes to ageCoeff(age)

LucasYoung:
This pull makes use of arrays indexed by age rather than a series of if-then statements to reduce the time complexity of the function to O(1).

I also made the arrays constants outside of the function rather than local variables. This may reduce the overall space usage of wilks.py, because we don't have to re-allocate variables each time the method is called.


sstangl:
Thank you! A very welcome improvement. That code was pretty bad!

It may be simpler still to put all the coefficients into a single array, filling in the missing cases with `0.00`. Then `ageCoeff()` could be a single table lookup in a `try` block, which would eliminate the subtractions.

PR #215, opened 2017-08-16
Remove hardcoding of year

LucasYoung:
In raw-probe the year is hardcoded as 2017, and displays a warning if the current year is not 2017. 
If there's a reason for this that I'm missing, no offense taken if this pull isn't merged!


sstangl:
The year hard-coding is intentional so that meets don't get lost when we transition from 2017 to 2018. For example, if it's January 2018, it's likely that meets will start getting posted only then from December 2017, so we still need to be checking the 2017 page.

One way to automate this would be to always check the meet pages from the current and previous year.

PR #221, opened 2017-08-20
some scripts I used

Borisdatzar:
@sstangl These are some of the scripts I used. I don't know if you want to add them all.


sstangl:
Can't hurt. They're small, and it's good to keep a record, even if they only work for you at the moment.

PR #222, opened 2017-08-21
Update lifters.csv

emp100k:
As per http://www.usapl.liftingdatabase.com/competitions-view?id=1013 Jade Dickens was lifting in Single-ply bench only as opposed to raw, the other ladies bench only categories were denoted with Raw M2, Raw T3, ect whereas Jade's category was only listed as M1.


sstangl:
Thanks. It looks like her name is also backwards in our version. They must have made an importation error, then fixed it after I'd already pulled it.

The USPA updates us when they make meet corrections, but the USAPL unfortunately does not.

emp100k:
Also on this one I was a bit concerned about  Cursti Cope's Bench press as well, as it also seemed unusually high for a raw bench press, and all her other meets have been done in Single-ply but the meet data at USAPL lifting database shows her in the Raw Bench Only T3 group so for now we have to take it at face value... 

    On Monday, August 21, 2017 3:27 PM, Sean Stangl <notifications@github.com> wrote:
 

 Merged #222.—

PR #223, opened 2017-08-23
Mo. State Powerlifting And Bench Press Championship

TrystanOakley:
@sstangl


sstangl:
Thanks :-)

TrystanOakley:
No problem! Sorry I didn't know if I should add him as bench only or not
since he did full power

PR #230, opened 2017-08-24
raw-probe uses probe library

LucasYoung:
I wrote a probe library to eliminate duplicate code in the probes. Doing this just for the raw-probe first so @sstangl can code-review probe.py before I move on to all the other probes. 💯


meet-data/probe.py
```diff
@@ -0,0 +1,41 @@
...
+def getenteredurls():
+    urls = []
+    for dirname, subdirs, files in os.walk(FEDDIR):
+        if 'URL' in files:
+            with open(dirname + os.sep + 'URL', 'r') as fd:
+                for k in fd.readlines():
+                    urls.append(k.strip())
+    return urls
+
+def main(federation, color, getmeetlist, URL):
```
sstangl: This probably shouldn't be called `main`... It's also doing too much because of the `color`, `getmeetlist`, and `URL` parameters.

It would probably be useful to just have this function take in a `meetlist` and an `enteredmeetlist`, then return a list of meets that haven't been entered. That's a little more complicated than it sounds, because URLs like "foo bar" and "foo%20bar" should compare equal. So they have to be canonicalized, somewhat.

The printing logic should probably be done in the federation's probe script itself, since not all federations have archives that need to be worked through.

meet-data/probe.py
```diff
@@ -0,0 +1,41 @@
...
+    print(msg, file=sys.stderr)
+    sys.exit(1)
+
+def gethtml(url):
+    with urllib.request.urlopen(url) as r:
+        return r.read()
+
+def getenteredurls():
+    urls = []
+    for dirname, subdirs, files in os.walk(FEDDIR):
```
sstangl: Instead of using a global, this should take `feddir` as an argument.

meet-data/probe.py
```diff
@@ -0,0 +1,41 @@
...
+from bs4 import BeautifulSoup
+import datetime
+import os
+import shutil
+import sys
+import urllib.request
+
+FEDDIR = os.path.dirname(os.path.realpath(__file__))
+
+def error(msg):
```
sstangl: This should probably be renamed `die`. I realize that I'm the one that named it!

meet-data/probe.py
```diff
@@ -0,0 +1,41 @@
+#!/usr/bin/env python3
+# vim: set ts=8 sts=4 et sw=4 tw=99:
+#
+# Library for probes for new meets
+
+from bs4 import BeautifulSoup
```
sstangl: BeautifulSoup isn't required here. Nor datetime. Nor, I think, shutil.

meet-data/raw/raw-probe
```diff
@@ -4,41 +4,30 @@
 # Probes for meet results from the 100%RAW Federation (RAW).
 # RAW posts meets to a page separated by year.
 
-
 from bs4 import BeautifulSoup
 import datetime
 import os
 import shutil
 import sys
+sys.path.append('../')
```
sstangl: Honestly, I'm not sure how best to do this in Python. I've just been using symlinks all over the place.

Could you research and find out what best practices are here? I bet they involve using a virtualenv... but I'd like to avoid that if possible.

meet-data/raw/raw-probe
```diff
@@ -60,27 +49,5 @@ def getmeetlist(html):
...
-    known = getenteredurls()
-
-    for m in meetlist:
-        if not m in known:
-            print(color('[RAW] ') + m)
-    print(color('[RAW] ') + "Continue working through archive.")
-
 if __name__ == '__main__':
-    main()
+    probe.main('[RAW]', color, getmeetlist, URL)
```
sstangl: The `main()` function should definitely remain in this file. Helper libraries shouldn't control `main`.

meet-data/raw/raw-probe
```diff
@@ -4,41 +4,30 @@
...
-    with urllib.request.urlopen(url) as r:
-        return r.read()
-
-
 def getmeetlist(html):
+    # URL needs updating every year.
+    # Also be careful to check if there are any result links that don't contain
+    # the string "results", since they won't get picked up by the script below.
+    if datetime.datetime.now().strftime("%Y") != "2017":
+        print("Warning: RAW fetch URL needs updating for new year.", file=sys.stderr)
```
sstangl: With `main` moved to the top, this can move back to its old location.

meet-data/raw/raw-probe
```diff
@@ -60,27 +49,5 @@ def getmeetlist(html):
...
-def main():
-    html = gethtml(URL)
-    meetlist = getmeetlist(html)
-
-    known = getenteredurls()
-
-    for m in meetlist:
-        if not m in known:
-            print(color('[RAW] ') + m)
-    print(color('[RAW] ') + "Continue working through archive.")
```
sstangl: The printing logic should probably remain in this file.

Although there's some additional logic to the printing that would probably be nice to have in the helper library -- specifically, some scripts accept a `--quick` argument that changes their printing behavior a little. It would be nice to standardize the usage of that argument.

LucasYoung:
@sstangl I addressed all of the changes you requested. Can you review the code again?

meet-data/raw/raw-probe
```diff
@@ -9,10 +9,16 @@ import datetime
 import os
 import shutil
 import sys
-sys.path.append('../')
+sys.path.append('scripts')
```
sstangl: I think this would prevent executing the probe from the federation directory. Maybe we want both "../../scripts" as well as "scripts"? Or the Makefile could set $PYTHONPATH itself?

meet-data/raw/raw-probe
```diff
@@ -49,5 +50,17 @@ def getmeetlist(html):
...
+    known = probe.getenteredurls(FEDDIR)
+    unknown = probe.getunknownmeets(meetlist, known)
+
+    for m in unknown:
+        print(color('[RAW] ') + m)
+
+    print(color('[RAW] ') + "Continue working through archive.")
+
 if __name__ == '__main__':
-    probe.main('[RAW]', color, getmeetlist, URL)
```
sstangl: GitHub is warning that newlines need to be added to the end of the files.

sstangl:
Thanks! I'll make a follow-up commit to make sure that ./raw-probe works from the federation directory, not just from the Makefile.

PR #236, opened 2017-08-26
Update social-media.csv

TrystanOakley:
Added the only missing instagram for the invitational BOB4 lifters I could find, Adam Werner @sstangl


sstangl:
Awesome, thank you.

PR #246, opened 2017-08-30
Age filter

LucasYoung:
This script is a precursor to filtering by age divisions on the website. It found 3547 different formatting of divisions, so this task is going to require a lot of manual work.


sstangl:
Thanks! Will help with #57. We need something for divisions like `lifter-data/name-corrections.dat`.

PR #252, opened 2017-08-31
Weight class selection responds to whether kg or lb is selected

LucasYoung:


web/scripts/common.js
```diff
@@ -1,12 +1,17 @@
...
 var common = (function () {
 
 return {
     kg2lbs: function(kg) {
-        return Math.round(kg * 2.20462262 * 10) / 10;
+        return Math.round(kg * KG_CONVERSION * 10) / 10;
+    },
+
+    lbs2kg: function(lb) {
+        return Math.round(lb / KG_CONVERSION *10) / 10;
```
sstangl: The weight classes have more-or-less set names that have nothing to do with their actual values -- for example, all of the weight classes in pounds are usually reported without any decimal component whatsoever: http://www.purepowerlifting.com/New/New%20Weight%20Classes

web/scripts/main-index.js
```diff
@@ -127,8 +127,26 @@ function makeDataProvider() {
     };
 }
 
+function generateWeightClasses() {
+    for(var i = 0; i < selClass.options.length; i++) {
+        var option = selClass.options[i];
+        var num = option.text.replace(/[^0-9\.]/g, '');
+        console.log(num);
```
sstangl: Some logging got left in here and below.

web/scripts/main-index.js
```diff
@@ -127,8 +127,26 @@ function makeDataProvider() {
...
+        if(option.text.indexOf('lbs') >= 0 && selWeightType.value === "kg") {
+            option.text = option.text.replace(num, Math.round(common.lbs2kg(num)));
```
sstangl: Is this logic a really fancy way to get around the "+" sign? :-)

Instead of doing all this replacing, it would probably be easier to just remember whether there was a "+" sign, and put it back after, always overwriting all the text.

web/templates/weightclass.frag
```diff
@@ -1,41 +1,41 @@
...
-    <option value="t82.5">-82.5</option>
-    <option value="t90">-90</option>
-    <option value="t90+">90+</option>
-    <option value="t100">-100</option>
-    <option value="t110">-110</option>
-    <option value="t125">-125</option>
-    <option value="t140">-140</option>
-    <option value="t140+">140+</option>
+  <optgroup id="traditional" label="Traditional">
+    <option value="t44">-44 kg</option>
```
sstangl: I don't think these should say kg/lbs. Instead of threading the state through the value, could we just have a JS global that remembers whether the selector is in kg or lbs?

web/scripts/main-index.js
```diff
@@ -127,20 +127,35 @@ function makeDataProvider() {
...
 function generateWeightClasses() {
-    for(var i = 0; i < selClass.options.length; i++) {
-        var option = selClass.options[i];
-        var num = option.text.replace(/[^0-9\.]/g, '');
-        console.log(num);
-        if(option.text.indexOf('lbs') >= 0 && selWeightType.value === "kg") {
-            option.text = option.text.replace(num, Math.round(common.lbs2kg(num)));
-            option.text = option.text.replace('lbs', 'kg');
+    if(selWeightType.value === "lb") {
+        for(var i = 1; i < selClass.options.length; i++) {
```
sstangl: Should this be `<=`? Also below. If `selClass.options.length == 1` then the loop will never execute.

Instead of duplicating the loop, it might be better to just use `selWeightType` to set a local variable, for example `var weightclasses = (selWeightType.value === "lb" ? LB_CLASSES : KG_CLASSES);`. Then there only has to be one loop, which is a little nicer.

PR #253, opened 2017-08-31
Nasa 2016

Borisdatzar:
@sstangl hey, i screwed up a bit here, didn;t realise that 1601 - 1604 already existed. I think they are different to my 1601-1604. My git commands are a bit rusty to fix this. Perhaps just allow this change? Whatever 16* were already there, I am getting to all of them within a few days anywa.


Borisdatzar:
@sstangl Please check this. I am not sure why its complaining about meet-data/nasa/nasa-standardize-csv whatever I am pushing now, is the latest version of that file.

sstangl:
Why are the scripts being pushed as part of this commit anyway?

It looks like your branch has diverged from master. You can fix that by running `git rebase -i master`, then following the instructions. You can use `git status` to figure out what files need to be fixed.

Borisdatzar:
@sstangl sorry for freaking out. I think this is right now. Ill push the script separately

meet-data/nasa/1613/1609/1604/check_lifters
```diff
@@ -0,0 +1 @@
+Summary: 0 errors, 0 warnings.
```
sstangl: This file looks like an error.

Borisdatzar: @sstangl Yes the entire subfolder 1609 should be deleted. What is the best way to do this so that I don't undo the changes you made to my commit?

Borisdatzar:
@sstangl i deleted the random wrong files

PR #256, opened 2017-09-01
Update lifters.csv

emp100k:
Seems like a pretty simple mistake, says he's in the -198 class but them lists his weight as 89.9, was probably supposed to be 189.9 and the 1 was misplaced.  
189.9/2.2046 = about 86.14, that said it is also possible that 89.9 was his metric weight, as it is just under 90kg (198lb) It would be strange if they notated it that way as the rest of the results have the bodyweights in pounds...


sstangl:
Thanks! The most likely thing seems to be that the `1` was dropped. SPF isn't too comfortable with using kilos.

PR #264, opened 2017-09-04
Convert main-meet and main-meetlist to TypeScript

LucasYoung:
Took me some time to get around to doing this! I held off off on common.js and database.js because @sstangl mentioned that they can incorporate some interfaces for opldb and meetdb.


sstangl:
Thanks! 👍

Issue #158, opened 2017-07-28
Add a link to the instagram and facebook page on the site


TrystanOakley:
I was thinking it would be beneficial to have a link to the sites instagram and facebook page on the main site somewhere

sstangl:
Oh, yeah, that's probably a good idea. Any ideas for where to put them?

TrystanOakley:
I know some sites have little F and I logos of the facebook and instagram. We could put those up in the top right hand corner to the left of the rankings link? Have it set  up so if you click on it itll take you to our social media? Or we could have another link in that corner that says social media and when you click it it takes you to another page that has links to our instagram and facebook page.

sstangl:
Logos are a great idea. Probably also something on the Contact page.

TrystanOakley:
That would work well

sstangl:
Closed by d324872c5a9bcbbfb8d6c37e2f0623ae62f7311d.

PR #276, opened 2017-09-08
Add Emily Lin and social media links

emilyytlin:
- Add Emily Lin to social-media 
- Add links to ig and fb in header and in contact page #158


sstangl:
Thank you! As a follow-up, it would be useful to replace `web/images/ig-glyph-logo_May2016.png` with the FontAwesome vector version as it occurs in the table.

FontAwesome weighs around 30kB, and the image weighs around 11kB, but both are pretty acceptable.

PR #280, opened 2017-09-09
UI enhancements

emilyytlin:
- Add sass
- Remove instagram logo to use font awesome
- Reorganize header - links to the left, metadata on the right
- Add `header.frag` for shared html header attributes


.gitignore
```diff
@@ -7,3 +7,4 @@ meet-data/uspa/fetch/
 meet-data/wrpf/fetch/
 web/node_modules/
 web/npm-debug.log
+*.scssc
```
sstangl: Newline needed at end of file.

web/Makefile
```diff
@@ -48,6 +48,7 @@ scripts: builddir npminit
 	    | ./try-uglify ${BUILDDIR}/scripts/jquery-slickgrid-2.2.6.min.js
 
 css: builddir
+	sass css/style.scss css/style.css --style compressed
```
sstangl: I don't have the command `sass` on my system. I could install it from my distribution's package manager, but there is a version of sass in npm as `node-sass`. Could we add that to the `web/package.json` `devDependencies` list, and use that version instead? Then nobody has to worry about whether or not they installed a local copy.

web/css/style.css
```diff
@@ -1,289 +1 @@
...
...333333;background-color:#3E4042;color:white;font-size:10pt;width:200px}button{padding:3px 8px 3px 8px;margin-left:5px;border:1px solid #333333;border-radius:4px;background-color:#3E4042;color:white;font-size:10pt;width:70px;cursor:pointer}button:hover{background-color:#56585C}button:active{background-color:#303133}input[type="checkbox"]{display:none}input[type="checkbox"]+label span{display:inline-block;width:19px;height:19px;margin:0px 1px 3px 5px;vertical-align:middle;background:url(images/check_radio_sheet.png) left top no-repeat;cursor:pointer}input[type="checkbox"]:checked+label span{background:url(images/check_radio_sheet.png) -19px top no-repeat}#lifter{font-size:10pt;margin-left:4px}div.topbar{position:fixed;width:100%;background-color:#1e1e1e}div.content{padding-top:82px}div.text-content{padding:82px 20px 20px 20px;max-width:800px;text-align:left;font-size:10pt}.downloadbar{display:flex;flex-direction:row}.downloaditem{border:1px;border-color:lime;border-style:solid;padding:10px;margin-right:10px}.nav{padding-right:10px}.nav__link+.nav__link{margin-left:10px}.nav__link--active{color:#569cd6}li{margin-bottom:10px}td.meetdivisiontext{color:#569cd6}.instagram{margin-left:8px}
```
sstangl: This file is now generated and should not be in version control. (It should be generated into the `web/build/` directory.)

web/css/style.scss
```diff
@@ -0,0 +1,272 @@
+$background-color: #1E1E1E;
+$alt-color: #242424;
```
sstangl: I'm not keen on the color changes. The original colorscheme was part of an intentional branding decision and shouldn't be changed by code without discussion. I'll address this in the general review message.

This patch is too large and needs to be broken up into smaller pieces. (If you no longer have the pre-squeezed commits, that's OK -- but if you do, even better.)

web/gen-header
```diff
@@ -3,75 +3,49 @@
...
+    active = 'nav__link--active' if activetab == tab else ''
+    print('<a href="{0}.html" class="nav__link {1}">{2}</a>'.format(tab, active, title))
 
+def main(activetab):
+    assert activetab in ['rankings', 'lifters', 'meet', 'meetlist', 'data', 'contact']
     print('<div id="header">')
-    print('OpenPowerlifting')
-
     print('<div class="nav">')
+    print('<a href="index.html" class="nav__link brand">OpenPowerlifting</a>')
```
sstangl: The CSS style of the `OpenPowerlifting` text changed somewhere, and it's much bigger than before.

web/gen-meettable
```diff
@@ -97,7 +98,7 @@ def makeinstagram(liftername):
 
     for row in SocialMediaCSV.rows:
         if row[nameidx] == liftername and row[igidx]:
-            return '<a href="https://www.instagram.com/%s"><img class="instagram" src="images/ig-glyph-logo_May2016.png"></a>' % (row[igidx])
+            return '<a href="https://instagram.com/%s" class="instagram"><i class="fa fa-instagram"></i></a>' % (row[igidx])
```
sstangl: This is great! The size of the IG logo is much smaller now, though, and harder to click.

web/scripts/common.ts
```diff
@@ -130,9 +130,8 @@ export function makeRowObj(row, index?: number): RowObject {
     if (socialmedia !== undefined) {
         let social = socialmedia[fullname];
         if (social !== undefined) {
-            name = name + ' <a href="https://www.instagram.com/' + social[0] + '">'
-                        + '<img class="instagram" src="images/ig-glyph-logo_May2016.png">'
-                        + '</a>';
+            name += ('<a href="https://instagram.com/' + social[0] +
```
sstangl: Probably here too with the glyph size.

web/templates/header.frag
```diff
@@ -0,0 +1,4 @@
+<title>OpenPowerlifting</title>
+<meta charset="utf-8">
+<link rel="stylesheet" type="text/css" href="https://cdnjs.cloudflare.com/ajax/libs/font-awesome/4.7.0/css/font-awesome.min.css">
+<link rel="stylesheet" type="text/css" href="style.css"/>
```
sstangl: File needs newline at end. (I think your editor can be configured to do this automatically.)

emilyytlin:
Updated

sstangl:
Thank you very much! Moving the navigation to the left is a big improvement.

There is a small amount of follow-up work, but I should be able to do that in separate commits rather than going through another review cycle.

Issue #277, opened 2017-09-08
Need a test script to check for invalid Instagram usernames.


sstangl:
Instagram unfortunately lets lifters change their Instagram usernames, which they do every so often. When they do that, all of our links break.

We should have a script that connects to Instagram for each of the usernames and looks for a 404. If a 404 is found, that lifter gets deleted from the database.

sstangl:
Ahem, that lifter's *Instagram handle* gets removed from the *social-media.csv* file. We don't delete all of their meet results out of spite :-)

emilyytlin:
Output:
```
emily@DESKTOP-RB4AVAL:/mnt/c/src/openpowerlifting$ time tests/check-instagram-exists lifter-data/social-media.csv
 Alexey Sivokon - sivokon - not found
 Andrew Le - poweredbynapalm - not found
 Ben O'Brien - b3nobrien8 - not found
 Cillian Hurley - bopper1985 - not found
 Collin Curtiss - colecurtissofficial - not found
 Daniel Casson - static_athlete2200 - not found
 Eva Dunbar - evalaurabelle - not found
 Jacob Welch - jacobwelch08 - not found
 Jesse Lanuevo - jesse__hc - not found
 John Rivas - johnrivas110 - not found
 Jonathan Harder - become_a_beast_181_2k - not found
 Jordan Lawler - chromedome106 - not found
 Justin Zottl - fullzottl - not found
 Mariella Fisher - _valkyrie_04 - not found
 Meicha Hardeman - yahwehsgeneral - not found
 Nancy Sarieh - fancynancypants23 - not found
 Odell Manuel - odellmanuelelitepowerlifter - not found
 Sheri Miles - smilesy_fitpoler - not found
 Steve Mann - smannofstell - not found
 Stina Akemalm - akemalm - not found
 Tom Davis - tdavis514 - not found
 Will Hunt - wj_hunt - not found
22 / 691 errors

real    11m40.366s
user    0m11.203s
sys     0m7.313s
```

PR #281, opened 2017-09-10
Check and fix invalid ig usernames #277

emilyytlin:


sstangl:
Thank you! The `check-instagram-exists` script looks good.

Some of the miscellaneous changes unfortunately break my meet-entering workflow. I think we should fix that before merging this. Another option would be to remove all the changes that aren't related to this bug.

The problem is that I (and other people) need to be able to execute the files in scripts/ and tests/ from any directory, when adding in meet results. To do that, we just symlinked the dependencies -- then the script doesn't need to know from where it was called (because importations will be done relative to the `PWD`.

Therefore we can't use something like `from scripts import foobar`, unless we add the directory containing `scripts` to the `PYTHONPATH`, as you did in the `Makefile`. But, that also doesn't work -- I start up a number of terminals, none of which inherit environment from each other. Additionally, it would be difficult to tell new volunteers to setup environment variables to make this easier, since they already have a hard time.

I'm not sure what the best solution is here. We've gone with symlinks and manual `sys.path.append('../../scripts')`. Each top-level script could also just figure out its file location and do the correct `sys.path.append` without hardcoding location too much.

The path of least resistance, at least for this patchset, is probably removing the parts unrelated to the issue.

emilyytlin:
Symlinks unfortunately do not work on my computer. I will undo these changes.

sstangl:
OK, that's good to know! I guess the `sys.path.append` logic works on your computer, because you used it in the other pull request. We could hardcode the directory relationship between the script location and the `scripts` directory by using `os.path.dirname(os.path.realpath(__file__))`.

emilyytlin:
Updated

sstangl:
Thank you! :-)

PR #289, opened 2017-09-12
Braggin Rights Push/Pull

TrystanOakley:
@sstangl apparently I skipped over so many wonky SPF meets I'm in 2014 now. Oops


sstangl:
Ah, this meet is one of those weird edge-cases.

The SPF clearly used kilos for the meets, but then converted back to pounds. When they converted back, they made rounding errors.

So you'll have to go through and round each entry to the nearest 2.5kg.

One easy way to do this is using LibreOffice Calc -- you can create a new column and use `=MROUND(column, 2.5)` and it will do the rounding for you. Save and re-load the file, and then you can delete the old un-rounded columns.

Or I can do it, either way.

PR #293, opened 2017-09-13
Update social-media.csv

TrystanOakley:
Added lifter Travis Paul Rose @sstangl


TrystanOakley:
Travis Failed because the name change I made isn't in effect yet so. It'll be ok

PR #298, opened 2017-09-15
Update lifters.csv

emp100k:
Michael Greeno's deadlift attempts were shifted over one, the subtotal of bench+squat was listed as the first attempt (380kg) the actual best deadlift was the third attempt from the results doc, 237.5kg.
Results doc also states the total was 617.5kg


sstangl:
Great catch, thank you!

PR #302, opened 2017-09-21
Update lifters.csv

emp100k:
342.5 kg (755) lb would be well over the current raw world record, sounds like this was supposed to be multi ply...


sstangl:
Good catch. Some other lifters were missed also.

PR #314, opened 2017-09-29
Summer Challenge

TrystanOakley:
@sstangl looks like a lot of misspelled names potentially. He also didnt label gender or equipment so. Another great SPF meet in the books.


sstangl:
Ah... we don't want to drop attempt information if a federation has provided it.

In this case, we need columns `Squat1Kg` through `Squat3Kg`, and same for `Bench1Kg` through `Bench3Kg` and the same for Deadlift.

I think that means that it needs to be re-imported... that sucks, but I can do it!

TrystanOakley:
Crud

TrystanOakley:
Did you already fix this?

sstangl:
Yes, the fixed commit was merged.

PR #335, opened 2017-10-18
Update social-media.csv

TrystanOakley:
Added lifters Leslie Evangelista, Alexis Callejo, Jenny Sutherland, and Shane Langston @sstangl


sstangl:
Thank you!

PR #350, opened 2017-11-06
Update lifters.csv

emp100k:
This appears to be a typo and was supposed to be 50kg not 150kg.  150kg not only would have ben well over the world record, but also 3 times her previous best bench press.


emp100k:
not sure why the check failed, all I did was delete the 1 in 150...

PR #363, opened 2017-11-14
nasa 1728

Borisdatzar:
@sstangl please review. I am not sure if this even qualifies as PL meet.


sstangl:
It does indeed qualify! Push/Pull meets are rarer than full meets, but they show up sometimes in a bunch of different federations.

PR #365, opened 2017-11-14
Update lifters.csv

TrystanOakley:
Changed Gaylon Black to Perseus Black per his request. He officially changed it on his membership card with the federation as well so. @sstangl


sstangl:
Needs an entry in the name change database also --

https://github.com/sstangl/openpowerlifting/blob/master/docs/lifter-name-corrections.md

Also uspa/0858.

TrystanOakley:
I changed it in both meets, would you still like me to add it to he name corrections since I changed it in all of his meets and he said he will be using that name from now on?

TrystanOakley:
Okay I’ll add it. He’s only done the two meets though. I verified it with
him.

PR #367, opened 2017-11-14
Update social-media.csv

TrystanOakley:
Added Perseus Black @sstangl


TrystanOakley:
The check will fail because I just now changed his name in the database and it hasn't been merged

PR #384, opened 2017-11-26
Update entries.csv

emp100k:
fixed bench-press weight 275 is way out of line for bench press especially for a 13 year old, likely a the decimal point was misplaced on entry.


emp100k:
oops, mathed wrong

PR #387, opened 2017-11-27
Update social-media.csv

TrystanOakley:
I found his instagram but he spells his last name as kravtcov instead of kravtsov like we have listed in our records. Is this just a translation issue? The instagram page looks like it would be the top ten bencher. @sstangl


sstangl:
"Kravtsov" is the correct transliteration of the Russian. In no context is "ц" ever written "tc".

I assume the error comes because he doesn't know the Latin character set, and kind of guessed. In Cyrillic, the character "c" is pronounced "s", so he must have gotten confused.

TrystanOakley:
Ahh okay I see. I figured you would know so thats why I asked.

PR #412, opened 2017-12-05
Canterbury Powerlifting meets

matt-pearce1:


sstangl:
Thank you very much! I hope the CSV formats were not too difficult to figure out.

There were some minor errors: `events.csv` should be `entries.csv`; the TotalKg for Josh Pierson was incorrect, and it was missing a Place column (easily auto-calculated by `scripts/calc-place`).

I wasn't able to find these results on the Facebook page of the federation, though. My big concern is that this Facebook page setup will make it extremely difficult for us to tell if we're missing NZPF meets, because there's no official list. It basically requires someone to either manually check their Facebook page constantly, or for NZPF lifters to send in meets manually.

matt-pearce1:
Thanks for fixing the mistakes I had in that file. I agree with you on the difficulty of completeness when dealing with results published on facebook. All regional organisations but Canterbury Powerlifting & Central Districts Powerlifting publish results on their websites, so most NZPF meets shouldn't be too much of an issue. Canterbury Powerlifting publish yearly calenders on their website and Central Districts used to(although their website is currently broken), so that should mitigate some of the problem.

I was planning on writing a web scraper to check their respective Facebook pages for things that look like meet results and send me the files if it finds them, although long term a better needs to be found.

PR #418, opened 2017-12-06
Add documentation for project setup

MichaelVessia:


sstangl:
Thanks! This will be a handy thing to point people toward.

PR #475, opened 2017-12-27
Added Otago NZPF meets and fixed some of the sexes that were obviously wrong in other files.

photon-rain:


sstangl:
Added without issue -- just needed to run the name standardization script. Thank you! 👍

PR #484, opened 2017-12-28
Removed duplicate PA meet.

photon-rain:
PA 1659 and PA 1660 were the same meet.


sstangl:
Thanks. I fixed it using a smaller patch with less directory motion -- because the order doesn't matter, and because I'm pretty sure no external websites have direct links to specific meet pages, I just deleted `pa/1660` and moved `pa/1676` to its position.

photon-rain:
Thanks Sean!

Issue #483, opened 2017-12-28
A lot of Australian Name corrections from James Wakefield


TrystanOakley:
> G'day guys, some Australian corrections for you:
> 
> * Lyndal Vile should be Lyndall Vile, she has one meet (Arnold Classic ProRawEight) under the incorrect name
> 
> * Patrick Bud Morrison should be Patrick Morrison, his one meet under the incorrect name is ProRawSix
> 
> * Rhys Keane should be Rhyss Keane, his one incorrect meet is ProRawNine
> 
> * Aaron Simms should be Aaron Simm, his two incorrect meets are ProRawNine and QLD States 2015 (GPC federation)
> 
> * Annabel Rendle-Short, Annie Rendle-Short, Annabelle Rendle-Short and Annie Short are all the same person, she now goes by Annie Short.
> 
> If you need independent confirmation for this, Markos Markopoulos (GPC and ProRaw president) should be able to easily do so.

The guy who sent us these corrections is James Wakefield on our Facebook Page

TrystanOakley:
https://www.facebook.com/james.wakefield.16

Issue #478, opened 2017-12-27
Add RPS meet for lifter Nicholas Brooks


TrystanOakley:
http://meets.revolutionpowerlifting.com/results/2017-meet-results/2017-rps-texas-merry-gainzmas/

photon-rain:
Added this meet.

sstangl:
This meet was already in the database as `rps/1779`.

The most likely scenario is that Nicholas Brooks competed in sleeves, was marked as Wraps because RPS, sorted by "Raw", didn't see his name, and then assumed we were missing data.

TrystanOakley:
That’s weird, I couldn’t find his meet anywhere. Well thank you. My bad.
Yes he was in sleeves.

Issue #410, opened 2017-12-05
Lifter Request to add WRPF meet


TrystanOakley:
I have no idea where the WRPF results are and I know they suck at posting them. This is what the lifter sent me.

Name: Edan Davey IG: https://www.instagram.com/edandavey/ Best total: 905 Fed: WRPF Australia Meet: WRPF Downunder Classic 2017 (Sept 02 2017)

He wants that meet added. I asked him to help me find the link to the meet results. He is currently looking. Will update if he finds it.

TrystanOakley:
http://australia.worldrawpowerlifting.com/competition-rules-results/results/2017-wrpf-down-under-classic/

TrystanOakley:
Its a screenshot...he already apologized lol

photon-rain:
Have added this meet.

PR #485, opened 2017-12-28
Name corrections for #483 and added meets for #410,#478

photon-rain:


sstangl:
Imported as https://github.com/sstangl/openpowerlifting/commit/1bb55656ee097c527088ee23754d17beecec8be8.

Thank you very much for importing that WRPF-AUS meet! Those ones are quite difficult. I created a new federation for them pathed `meet-data/wrpf-aus/`, since they post results independently of the main Russian `wrpf` instance.

The report of a missing RPS meet was unfortunately an error -- the meet was already in the database, but the lifter was unable to find it.

PR #512, opened 2018-01-05
Update results.csv

emp100k:
Christine Woords total was put in the best deadlift column, fixed.


sstangl:
This commit accidentally edited the `results.csv` instead of the `entries.csv`.

PR #516, opened 2018-01-05
Update entries.csv

emp100k:
Deadlifts for Tyler Sebeart and Lance Gray appeared wildly out of line with their squat and bench numbers (as well as being a massive world record at 220 and national record at 242)  This seemed extremely suspicious as they are both novice lifters I've never heard of before.  Judging by their squat and bench numbers the 4s in the deadlift most likely should have been ones.


sstangl:
Thanks. That is indeed what the federation reported, but your change is probably correct.

PR #518, opened 2018-01-05
Add FFForce Championnats de France Elite 2015-2016 & standardize names

manvari:
Names were standardized for a few French lifters.


sstangl:
Thank you! I did not review the data carefully: I just made sure that the formats looked correct. You are trusted, use that power for good :-)

PR #534, opened 2018-01-08
Scripts

Borisdatzar:
@sstangl small script changes

### DIFF
--- a/meet-data/ipf/ipf-standardize-csv
+++ b/meet-data/ipf/ipf-standardize-csv
@@ -4,7 +4,7 @@
 # Standardize the results.csv to the OpenPowerlifting
 # internal format.
 #
-
+import codecs
 import sys
 from oplcsv import Csv
 
@@ -124,11 +124,15 @@ def unreverse_names(csv):
 
         row[nameidx] = name
 
-
+def hunt_dq(csv):
+    plidx = csv.index('Place')
+    for row in csv.rows:
+        if row[plidx] == '':
+            row[plidx] = 'DQ'
 def main(filename):
     csv = Csv()
-    with open(filename, 'r') as fd:
-        csv.rows = [x.strip().split(',') for x in fd.readlines()]
+    with codecs.open(filename, 'r', encoding="utf-8",  errors='ignore') as fd:
+        csv.rows = [x.strip().replace("?","").split(',') for x in fd.readlines()]
 
     # Fieldnames given for the detailed spreadsheet.
     fieldnames = []
@@ -189,6 +193,7 @@ def main(filename):
 
     strip_whitespace(csv)
     unreverse_names(csv)
+    hunt_dq(csv)
 
     csv.write(sys.stdout)
 
--- a/meet-data/nasa/todo/getpdfs.sh
+++ b/meet-data/nasa/todo/getpdfs.sh
@@ -3,16 +3,16 @@
 #One by one, it fetches the pdf, creates a new folder, moves the pdf to that folder, applies tabula to it
 #runs nasa-standardize to it, applies check-entries and stores the output.
 input="pdf_small"
-COUNTER=10
+COUNTER=1
 while IFS= read -r var
 do
-  mkdir /Users/blerner/openpowerlifting/meet-data/nasa/todo/"$COUNTER"
+  mkdir /Users/blerner/openpowerlifting/meet-data/nasa//140"$COUNTER"
 	wget $var
-	mv *.pdf /Users/blerner/openpowerlifting/meet-data/nasa/todo/"$COUNTER"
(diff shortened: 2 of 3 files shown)


meet-data/nasa/todo/getpdfs.sh
```diff
@@ -3,16 +3,16 @@
 #One by one, it fetches the pdf, creates a new folder, moves the pdf to that folder, applies tabula to it
 #runs nasa-standardize to it, applies check-entries and stores the output.
 input="pdf_small"
-COUNTER=10
+COUNTER=1
 while IFS= read -r var
 do
-  mkdir /Users/blerner/openpowerlifting/meet-data/nasa/todo/"$COUNTER"
+  mkdir /Users/blerner/openpowerlifting/meet-data/nasa//140"$COUNTER"
```
sstangl: Three comments:

1. The `//` here is redundant, likely a typo.

2. It looks like this "140" is repeated in a bunch of places. Instead of repeating it, it would be better to use a variable.

3. "140" isn't correct, because `${COUNTER}` will be >= 10 for 2014. That will produce names like `14010` and `14011` instead of `1410` and `1411`. Really what you want to do is prepend `14` and left-zero-pad `${COUNTER}`.

meet-data/nasa/todo/getpdfs.sh
```diff
@@ -3,16 +3,16 @@
...
 	wget $var
-	mv *.pdf /Users/blerner/openpowerlifting/meet-data/nasa/todo/"$COUNTER"
-	cd /Users/blerner/openpowerlifting/meet-data/nasa/todo/"$COUNTER"
+	mv *.pdf /Users/blerner/openpowerlifting/meet-data/nasa/140"$COUNTER"
+	cd /Users/blerner/openpowerlifting/meet-data/nasa/140"$COUNTER"
 	for f in *;
 	do
-		java -jar /Users/blerner/openpowerlifting/tabula-1.0.1-jar-with-dependencies.jar -l $f >"results.csv"
+		java -jar /Users/blerner/openpowerlifting/tabula-1.0.1-jar-with-dependencies.jar -l -p all $f >"results.csv"
```
sstangl: My Java does not have a `-p` option. On my system, this will fail. What does it do on your system? Can you maybe make it MacOS-only?

Granted, this is not such a big deal, since your username is hardcoded all over the place above :-)

meet-data/ipf/ipf-standardize-csv
```diff
@@ -124,11 +124,15 @@ def unreverse_names(csv):
 
         row[nameidx] = name
 
-
+def hunt_dq(csv):
+    plidx = csv.index('Place')
+    for row in csv.rows:
+        if row[plidx] == '':
+            row[plidx] = 'DQ'
 def main(filename):
```
sstangl: Nit: there should be two lines of vertical whitespace between functions.

meet-data/ipf/ipf-standardize-csv
```diff
@@ -124,11 +124,15 @@ def unreverse_names(csv):
...
     csv = Csv()
-    with open(filename, 'r') as fd:
-        csv.rows = [x.strip().split(',') for x in fd.readlines()]
+    with codecs.open(filename, 'r', encoding="utf-8",  errors='ignore') as fd:
+        csv.rows = [x.strip().replace("?","").split(',') for x in fd.readlines()]
```
sstangl: Do not silently remove errors. The "?" are important, because they mark places that you will have to fix manually. This is particularly important for names -- we can't just silently drop random letters because of encoding errors.

Borisdatzar: Problem is I am getting a lot more "?" than you. without this line, the script actually fails.

meet-data/nasa/todo/getpdfs.sh
Borisdatzar: Ye this whole script is just for me really....

meet-data/nasa/todo/getpdfs.sh
Borisdatzar: the -p is very importnat for me. It stands for "pages". Without -p all only the first page gets done, which happened in the past

meet-data/nasa/todo/getpdfs.sh
sstangl: Even if the script is just for you, it still seems like an odd logical error.

meet-data/ipf/ipf-standardize-csv
sstangl: Is there a way to remove the "?" more selectively, so that names don't get mangled?

sstangl:
Closing for the moment. Can re-open.

PR #544, opened 2018-01-09
IPF 1705

Borisdatzar:
@sstangl plz rev


meet-data/ipf/1705/URL
```diff
@@ -0,0 +1 @@
+results_f.csv
```
sstangl: This does not look like the right URL :-)

sstangl:
Looked good, thanks! Three minor changes:

1. The `URL` file didn't contain the URL.
2. The `results_m.csv` and `results_f.csv` files were concatenated into a single `results.csv`.
3. The meet name should be as the federation specified, including the word "Powerlifting". In non-IPF federations it's OK to omit that, but the IPF is special.

PR #545, opened 2018-01-09
nasa 1706

Borisdatzar:
@sstangl plz rev


sstangl:
Thanks! I changed it to just be a single `results.csv` by concatenating the other results files.

PR #573, opened 2018-01-14
Update how-to-add-a-meet.md

camdenmariotti:
Changed the `bodyweight` column to if reported.


sstangl:
Thanks! I also noticed that when you create files in Windows, and transfer them to Ubuntu, they're in slightly-wrong formats (namely, Windows format). You can fix that by running the following commands on each file:

- `chmod -x ${FILE}` (this makes the file non-executable -- all Windows files are executable for some reason.)
- `dos2unix ${FILE}` (this changes the line endings from `\r\n` (Windows format) to `\n` (Linux format / the normal format)).

PR #639, opened 2018-01-25
Added meet results from a 1993 NZPF magazine

photon-rain:


sstangl:
Looks good. Thank you!

PR #640, opened 2018-01-25
Add 2015 APF Gulf Coast Winter Bash.

MichaelVessia:
Check will fail on a few lifters who had multiple entries, as they were in multiple divisions.  Not sure how to handle those so I'm passing it along


sstangl:
Thank you, looks good! And thanks for fixing the Division errors. I think the commits were made on your `master`, so you may want to blow that away since I merged a squashed commit.

MichaelVessia:
Yep I forgot to branch. Thanks

PR #652, opened 2018-01-26
Added ProRaw 1-3 and seperated Martin Nguyen's

photon-rain:


sstangl:
Great, thank you! The `URL` files for ProRawTwo and ProRawThree appear to link to a site that is behind a login form. That means that the results are not crawlable by the Internet Archive. Would you please follow up this commit with some `results` file for those two meets, so that we can have a record of the original?

photon-rain:
Sure thing -I'll add them now

Issue #290, opened 2017-09-12
NZPF's Oceania Championships may be OceaniaPF


sstangl:
NZPF hosted the meets, but it looks like a bunch of them were jointly hosted by AsianPF and OceaniaPF.

photon-rain:
Fixed!

PR #659, opened 2018-01-26
Fixed OceaniaPF results that were marked as NZPF

photon-rain:


sstangl:
@photon-rain Oh dear. Could you please resubmit this, just moving a single meet to replace the NZPF meet that got knocked out?

For example, if the NZPF meet was `1602`, and the last meet beginning with `16xx` was `1626`, move `nzpf/1602` -> `oceaniapf/something` and `nzpf/1626` -> `nzpf/1602`.

Since the order is arbitrary, that accomplishes the same thing but with less churn.

sstangl:
Note that this is intended to fix #290.

photon-rain:
There were 6 NZPF meets that got knocked out - the Oceania champs were listed as 6 different meets (Raw, Single-ply & Bench only with Men's and Women's results for each).

sstangl:
Oh, dang. OK!

PR #667, opened 2018-01-27
British Classic bench 2015

JoWhiteley:


sstangl:
This one was added, but it contained significant errors. Many of the entered lifts and `BestBenchKg` were incorrect, all failed lifts were missing, and some decimal points were misplaced. The `Division` information was frequently erroneous, with many `M2` lifters entered as `Open`. There were also some name misspellings.

I checked it very carefully from the source document and think I caught most things. Unfortunately I think any remaining errors would be pretty subtle, but hopefully review caught them all!

PR #675, opened 2018-01-29
Add Dockerfile and Docker docs

chris-vest:


sstangl:
Heck yes!

Is it possible to maintain two Dockerfiles in the same project, for example for the `server/` component (which will eventually replace `web/`)?

PR #687, opened 2018-02-01
Update entries.csv

TrystanOakley:
Alaina Hernandez was sleeved at this meet @sstangl


sstangl:
When submitting a correction, could you also please include the source? For example, a link to the Instagram post showing sleeves, or just "the lifter said so."

One of the benefits of using Git is that this information will be tracked for each meet -- so we can go back in time and see why we changed things.

PR #694, opened 2018-02-02
Old Canterbury lifts from the Wayback Machine

photon-rain:


sstangl:
Thanks. I didn't review line-by-line, but it generally looked right. Thanks for merging the Bench & Deadlift on that one meet.

PR #697, opened 2018-02-02
Added NZ Champs for 04-17 & South Island champs

photon-rain:
Ignore the silly amount of commits, still learning how to use branches properly.


photon-rain:
South Island Champs were supposed to be in a different branch- sorry about that.

sstangl:
Thanks! Some results files are missing, but not a big deal. That's quite a large number of meets, and all in different formats. I hope it's getting easier!

PR #699, opened 2018-02-02
Added remaining Equipped Champs

photon-rain:


sstangl:
Got it merged, but there were some number conflicts. So the meets may have different numbers now.

photon-rain:
Thanks for fixing the numbering!

PR #710, opened 2018-02-02
Update name-corrections.dat

emp100k:


sstangl:
@emp100k Those two names are identical.

PR #720, opened 2018-02-05
Update ipf-import

photon-rain:
Ipf-import now works for bench only meets & can combine male and female meets. meet.csv is now partially pre-filled.

### DIFF
--- a/meet-data/ipf/csv-cat
+++ b/meet-data/ipf/csv-cat
@@ -1,0 +1,1 @@
+/home/matt/openpowerlifting/scripts/csv-cat
--- a/meet-data/ipf/csv-cat
+++ b/meet-data/ipf/csv-cat
@@ -1,1 +1,0 @@
-/home/matt/openpowerlifting/scripts/csv-cat
--- a/meet-data/ipf/ipf-import
+++ b/meet-data/ipf/ipf-import
@@ -18,7 +18,7 @@ cd "$1"
 
 
 #One file
-if [ $# -nt 2 ];then
+if [ $# -eq 2 ];then
 	# Download the results page.
 	wget --output-document=results.html "$2"
 	cp results.html results.xls
--- a/meet-data/ipf/ipf-import
+++ b/meet-data/ipf/ipf-import
@@ -16,7 +16,6 @@ fi
 mkdir "$1"
 cd "$1"
 
-
 #One file
 if [ $# -eq 2 ];then
 	# Download the results page.
@@ -53,13 +52,16 @@ if [ $# -eq 2 ];then
 	# Failures are denoted as "X", when we just want a blank space.
 	sed -i -e 's/,X,/,,/g' results.csv
 
+	#Skips are sometimes denotes as "#"
+	sed -i -e 's/#//g' results.csv
+
+
 	# Disqualifications use ascii 0x97, a long dash.
 	sed -i -e 's/\x97/DQ/g' results.csv
 
 
 
 else #two or more files
-
 	for (( ii=2; ii<=$# ; ii++))
 	do
 		# Download the results page.
@@ -68,7 +70,7 @@ else #two or more files
 
 		# Save the website URL for future use, since it's pretty hard to tell
 		# what meets on the site are tracked and which aren't.
-		echo ${!ii}  > URL
+		echo ${!ii}  >> URL
 
 
 		# Replace any commas in the xls file with periods.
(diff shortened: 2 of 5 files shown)


meet-data/ipf/csv-cat
```diff
@@ -0,0 +1 @@
+/home/matt/openpowerlifting/scripts/csv-cat
```
sstangl: Hm... This symlink may need more relativism!

meet-data/ipf/ipf-import
```diff
@@ -8,49 +8,106 @@ set -e
...
+	sed -i -e 's/#//g' results.csv
+
+
+	# Disqualifications use ascii 0x97, a long dash.
+	sed -i -e 's/\x97/DQ/g' results.csv
+
+
+
+else #two or more files
+	for (( ii=2; ii<=$# ; ii++))
```
sstangl: Surely this can handle the `$#=2` case also?

meet-data/ipf/ipf-parse-post
```diff
@@ -7,13 +7,40 @@ SCRIPTDIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
...
+	${SCRIPTDIR}/csv-cat ${entries[@]} > entries.csv
+	rm ${entries[@]}
+
+
+fi
+
 
 # Show any warnings.
-${REPOTESTDIR}/check-entries-csv entries.csv
+#${REPOTESTDIR}/check-entries-csv entries.csv
```
sstangl: Should still be run.

meet-data/ipf/ipf-parse-post
```diff
@@ -7,13 +7,40 @@ SCRIPTDIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
...
-${SCRIPTDIR}/ipf-standardize-csv results.csv > entries.csv
+if [ -e results.csv ];then
+	# Use IPF-specific knowledge to parse results.csv into entries.csv.
+	${SCRIPTDIR}/ipf-standardize-csv results.csv > entries.csv
+else
+	n=`ls  | grep 'results.*\.csv' | wc -l`
+
+	for (( ii=1; ii<=${n};ii++))
+	do
+		${SCRIPTDIR}/ipf-standardize-csv "results"$ii".csv" > "entries"$ii".csv"
```
sstangl: If this is using `${SCRIPTDIR}`, then you shouldn't need the `csv-cat` symlink.

meet-data/ipf/ipf-standardize-csv
```diff
@@ -124,41 +153,106 @@ def unreverse_names(csv):
...
+def add_sex(csv,sex):
+    sexidx = csv.index('Sex')
+    for row in csv.rows:
+        row[sexidx]=sex
+
+
+#Finds the sex of the lifters in a meet from the weight classes
+def meet_sex(csv):
+    mens_classes = ['-53','-59','-66','-74','-83','-93','-105','-120','120+']
+    womens_classes = ['-43','-47','-57','-63','-72','-84','84+'] #Don't have -52 as this overlaps with an old class
```
sstangl: Older meets use the old system, where men and women have the same weightclasses. (The "Traditional" classes.)

meet-data/ipf/ipf-parse-post
photon-rain: Good point, removed this symlink.

meet-data/ipf/ipf-parse-post
photon-rain: Not sure why I commented this out, changed.

meet-data/ipf/ipf-standardize-csv
photon-rain: The old_mens_only and old_womens_only arrays are there for meets which use the old weight-classes. The old weight classes are still slightly different in that there are women's 44,48,90+ classes and no women's classes for 100,110,125,125+ - so lifter sex should still be able to be determined from the weight classes.

 Since the script doesn't know which era weight-classes it's dealing with I removed 52kg from the new women's weight classes - otherwise a -52kg men's class may have been flagged as a women's class.

meet-data/ipf/ipf-import
photon-rain: I've separated the $# =2 and $# > 2 cases just so that the results files can be labelled differently, although it does mean a lot of code is repeated for a small amount of functionality. I can change this if you want?

sstangl: If it's not too much work to avoid the repetition, that would be preferable.

photon-rain: Removed in the latest commit.

meet-data/ipf/ipf-import
```diff
@@ -16,97 +16,56 @@ fi
...
+fi
+
+
+#Process $#-1 meet files
+for (( ii=2; ii<=$# ; ii++))
+do
 	# Download the results page.
-	wget --output-document=results.html "$2"
-	cp results.html results.xls
+	wget --output-document="results"${r_nums[$((ii-2))]}".html"  ${!ii}
```
sstangl: Could use a temp variable instead of repeating `${r_nums[$((ii-2))]}`.

PR #722, opened 2018-02-05
2018 updated

JoWhiteley:
YNE Winter Open entries corrected


sstangl:
Thank you! The following changes were made:

1. `Division` was changed from `F-C` etc. to `S`,`J`,`M1`, etc. This is necessary to remember which lifters should be given age division rankings eventually -- since their ages were not explicitly reported.
2. A guest's `Place` should be `G` (instead of `Guest`).
3. When attempts are given, and all attempts were failed `BestSquatKg` and friends should be empty instead of negative. (A negative number is allowed there in case the attempts aren't available.)
4. The date order is `YYYY-MM-DD` (numerical sort order).

PR #727, opened 2018-02-06
Update entries.csv

emp100k:
In the results Sonia is listed as 84+ not 84 for weight class, also in her previous and next meet (only about a month and a half before and after this meet) her bodyweight is listed as nearly 50kg higher.


sstangl:
Great catch, thank you.

PR #730, opened 2018-02-07
ADFPA 9502

camdenmariotti:
Not sure why those previous meets are showing, but I uploaded `adfpa/9502`.


sstangl:
See PM -- this merge commit introduced 21 changes, and the same binary file twice. I would like to keep the history linear, with zero merge commits, and such that every commit builds. In the UI, this can be done by a "Squash Commit", which just uses `git rebase -i` to squash all the changes into a single one.

sstangl:
Also, `adfpa/9501/results.txt` appears to be a binary file, not a text file:
```
[sstangl@mazu 9501]$ file results.txt 
results.txt: PDF document, version 1.4
```
I'm going to rewrite the master branch to get these binary objects out of the repo... hopefully that will be OK! You'll have to force-pull.

camdenmariotti:
@sstangl Sorry for the confusion.  In the future I would like to use Ubuntu only when necessary and use GitHub for desktop and/or the website as much as possible. I am still trying to figure this stuff out and it is just a lot to try and process at the moment.

PR #735, opened 2018-02-08
Update entries.csv

emp100k:
Dion Smith is male not female, the bodyweight looks wrong too (listed as a 185 in an earlier meet, so 114 would be very low) but I can't divine his actual bodyweight/weight class from the results...


sstangl:
Closed as afc089b884228a7bac1f5881de68b.

PR #738, opened 2018-02-09
Added 3 new tests.

photon-rain:
Sex weight-class consistency, that attempts are ascending and that successful attempts are not retaken.

### DIFF
--- a/tests/check-entries-csv
+++ b/tests/check-entries-csv
@@ -689,24 +689,13 @@ def check_row(row, csv):
     check_row_lift_consistency(row, csv, 'BestDeadliftKg',
                                           ['Deadlift1Kg','Deadlift2Kg','Deadlift3Kg'])
 
-    if 'Squat4Kg' in csv.fieldnames:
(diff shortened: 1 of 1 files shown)


tests/check-entries-csv
```diff
@@ -624,6 +689,27 @@ def check_row(row, csv):
     check_row_lift_consistency(row, csv, 'BestDeadliftKg',
                                           ['Deadlift1Kg','Deadlift2Kg','Deadlift3Kg'])
 
+    if 'Squat4Kg' in csv.fieldnames:
+        check_row_lift_order(row, csv,['Squat1Kg','Squat2Kg','Squat3Kg','Squat4Kg'])
```
sstangl: Since `check_row_lift_order()` already checks for column existence, couldn't you just always call it with `Squat4Kg` in the list? If that column isn't present, it looks like it would just be ignored.

photon-rain: Good point - amended.

PR #739, opened 2018-02-09
Added 2016 Canterbury Raw Cup & fixed NZPF warnings

photon-rain:
Down to 1096 warnings from check-entries-csv now. Progress!


photon-rain:
Removed 53kg from IPF mens classes as the IDFPF has a 53kg womens class.

photon-rain:
Apparently SPF has 44kg & 48kg men's classes so removed those too. Seems like this check isn't going to be as useful as I thought - it still catches a lot of IPF mistakes though.

sstangl:
All those changes look plausible. I skimmed through until my eyes glazed over :shipit:

Thank you so much for doing all that work!

PR #742, opened 2018-02-09
EPA 2016 up to SE

JoWhiteley:


photon-rain:
Hey Jo which meet is 1631 supposed to be? The data doesn't seem to match up with the meet.

sstangl:
@photon-rain It looks like the results for 1631 are here:

http://www.britishpowerliftingsoutheast.co.uk/Results/Open%202016%20Sat.pdf
http://www.britishpowerliftingsoutheast.co.uk/Results/Open%202016%20Sun.pdf

JoWhiteley:
It should be the South East Open.  I'm not taking the results from the individual regional websites.  They are (or should all be) on www.englishpowerlifting.co.uk which is the overall English fed.  Let me check - I did loads today maybe I got distracted

photon-rain:
Ok I was just reading it wrong - the ordering being different threw me off for a minute.

photon-rain:
Does EPA not assign totals for non full-meet lifters or are those DQs?

JoWhiteley:
DQs

JoWhiteley:
Although there's something wrong there.  Those DQs don't make sense.  I will go back to the source tomorrow and figure it out.

sstangl:
It's OK to land things in a state that doesn't make sense and then just file an issue for it to be resolved later. The only real problem is losing track of things.

JoWhiteley:
OK cool.  I will log myself an issue to follow it up.

PR #747, opened 2018-02-09
Update entries.csv

emp100k:
Leslie Williams bodyweight is 74.2 kg, not 74.2 lb.


sstangl:
Good catch, thanks :-)

PR #755, opened 2018-02-10
Tests now run on windows.

photon-rain:
Removed dependence of symlinks and made files always open with utf-8 encoding.

This requires the installation of the python module colorama - is this fine with you Sean or do you want me to disable colours for Windows?


photon-rain:
Ok apparently this breaks Travis - not sure if this is a fixable problem?

PR #756, opened 2018-02-10
Epa 1631 fix

JoWhiteley:
Went back to the source document to fix some issues.  Also noticed some people marked as equipped that I had down as raw.


photon-rain:
Thanks Jo!

PR #759, opened 2018-02-11
Beau Moores' 600lb bench

photon-rain:
Added Beau Moore's 600lb bench press & the fed it was set in. Modified the server files to add the new fed.


sstangl:
I think they probably want to be called `RAWU` or `RawUnited`... `RUNI` looks like it's pronounced "Runny"!

PR #802, opened 2018-02-20
Update entries.csv

emp100k:
All of Rachels other meets are Multi-ply, where she has comparable (some are lower) bench numbers, this would be a ATWR if raw so I believe it is actually Multi-ply.


sstangl:
Indeed. Good catch, thank you!

PR #804, opened 2018-02-21
Added 2008 WABDL Worlds

camdenmariotti:
Was able to run a make check successfully with no critical errors.


photon-rain:
For future reference: DQs go in the Place column not the TotalKg column.

camdenmariotti:
Ok, I’ll get it right next time.

sstangl:
Thanks to both of you for adding these WABDL meets! You're tackling some really hard ones. I really appreciate the effort. :+1:

PR #818, opened 2018-02-24
Added   2008 WABDL National Collegiate Bench Press & Deadlift Championships

camdenmariotti:
Had to change `0801` to `0802`, since this meet occurred earlier in the year than WABDL Worlds (previously `0801`).


camdenmariotti:
Wow, I forgot to change LBS to KG and run `make check`. I'm going to do that and submit again.

sstangl:
The order of the meet numbers is unimportant. It should be the order in which they were entered. `0802` may have an earlier `Date` than `0801`.

camdenmariotti:
Oh, my bad. For some reason, I thought it was by date. Should I go ahead and fix that back or does it even matter?

PR #843, opened 2018-03-01
Fixed EPF data.

photon-rain:
Changed the name of EPF/1305 and relabelled lifters as SP, deleted EPF/1703, added mens data to EPF/ 1107.


photon-rain:
Just noticed the IPF files that shouldn't have changed -will fix those now.

Issue #607, opened 2018-01-21
Fix Sorting Issue for Total and Wilks


TrystanOakley:
When you sort by "Total" have it refer to "wilks" in the event of two people having the same total and give a higher place to the person with the higher wilks.

PR #844, opened 2018-03-01
Bugfix/607 sorting total wilks

MichaelVessia:
Fix Sorting Issue for Total and Wilks #607

Wilks will sort in the same direction as the sort for Total.


MichaelVessia:
Not sure why it added my commits from a previous PR.  Let me know if it is an issue.

web/scripts/database.ts
```diff
@@ -54,29 +54,51 @@ export function db_filter(indices: number[], rowcmpfn): number[] {
...
-        if (isNaN(av))
-            av = Number.MIN_VALUE;
-        if (isNaN(bv))
-            bv = Number.MIN_VALUE;
-        return bv - av;
+        let av1 = Number(opldb.data[a][col1idx]);
+        let bv1 = Number(opldb.data[b][col1idx]);
+
+        let av2 = Number(opldb.data[a][col2idx]);
+        let bv2 = Number(opldb.data[b][col2idx]);
```
sstangl: In the common case where `av1 !== bv1`, performing loads for `av2` and `bv2` are wasted operations and quite expensive. It would be good to only execute these loads in the `if (av1 === bv1)` path below.

web/scripts/database.ts
```diff
@@ -54,29 +54,51 @@ export function db_filter(indices: number[], rowcmpfn): number[] {
...
+        if (isNaN(av2))
+            av2 = extreme;
+        if (isNaN(bv2))
+            bv2 = extreme;
+
+        // Default comparison from the single-col sort
+        let comparison = minFirst ? (av1 - bv1) : (bv1 - av1);
+
+        // If the first column is equal, compare the second
+        if (av1 === bv1)
```
sstangl: In the common case of `av1 !== bv1`, performing the comparison was wasted work.

web/scripts/database.ts
```diff
@@ -54,29 +54,51 @@ export function db_filter(indices: number[], rowcmpfn): number[] {
...
+            comparison = minFirst ? (av2 - bv2) : (bv2 - av2);
+            
+        return comparison;
```
sstangl: Please change this to prefer multiple return points -- JavaScript JITs handle that better than single return points. (i.e., instead of storing `comparison`, just return the value directly.

PR #850, opened 2018-03-01
Update social-media.csv

Shang63:
Added instagram account for Shannon Graszat


sstangl:
Thanks! Should be on the site in short order (~5 minutes).

PR #857, opened 2018-03-03
IPF 2010-2017

photon-rain:
All IPF results from 2010-2017. Moved Arnold 2016 results to napf and usapl-archive folders.
Name-corrections.dat is included in a seperate pull as it contains many lifters not in this pull.


lifter-data/name-corrections.dat
```diff
@@ -2443,7 +2443,7 @@ Sam Rowland-Jones,Sam Rowland Jones
 Danny O'Brien,Danny Obrien
 Elijah May,Elijiah May
 Tatiana Brodetská,Tatiana Brodetska
-Ernő Eck,Erno Eck
+Ernö Eck,Ernő Eck,Erno Eck
```
sstangl: His name is Ernő Eck, just for reference. https://www.facebook.com/erno.eck

Issue #849, opened 2018-03-01
Missing IPF data


photon-rain:
IPF doesn't have complete data for 1988-1995 Men's Worlds and 1990-1995 Bench Worlds . Data is missing completely for World Disabled Bench 1999,1990-1995, 1999 Bench Worlds.  AllPowerlifting has most of this data so it exists somewhere, check magazines when they are scanned to see if these results are there.

PR #859, opened 2018-03-03
IPF 1990-1999

photon-rain:
Almost all IPF results from 1990-1999. Name-corrections.dat is included in a seperate pull as it contains many lifters not in this pull.


photon-rain:
This is only missing the meets listed in #849.

PR #860, opened 2018-03-03
IPF 1980-1989

photon-rain:
All IPF results from 1980-1989. Name-corrections.dat is included in a seperate pull as it contains many lifters not in this pull.


photon-rain:
Note that this give weight class warnings for the World Games. The weight class check will need to be reworked slightly to get rid of this warning.

PR #862, opened 2018-03-03
Name corrections file for IPF results

photon-rain:
Combined name corrections file for all IPF commits.


lifter-data/name-corrections.dat
```diff
@@ -78,12 +83,12 @@ Alexandra De L'Armour,Alexandra DeL'Armour,Alexandra De L'armour
 Alexandra Murphy-Jelley,Alexandra Murphy Jelley
 Alexandre Pelletier,Alexandre Pellletier,Alexand Pelletier
 Alexis Guzmán,Alexis Guzman
-Alf-Ivar Hellerud,Alf Ivar Hellerud
```
sstangl: The changes in this file remove a large number of existing name corrections. Was that intentional? If yes, why?

photon-rain: Not intentional, due to this pull being based on an out of date upstream. Will integrate the existing corrections.

photon-rain:
Added back the name corrections that were missing.

PR #870, opened 2018-03-04
NZPF Probe

photon-rain:
Created a probe for NZPF, added missing meets found by probe & did some cleanup of meet names.


meet-data/nzpf/nzpf-probe
```diff
@@ -0,0 +1,125 @@
+#!/usr/bin/env python3
+# vim: set ts=8 sts=4 et sw=4 tw=99:
+#
+# Probes for meet results from the Commonwealth Powerlifting Federation.
```
sstangl: :-)

PR #872, opened 2018-03-05
IPF meet name cleanup.

photon-rain:
Tidied up names of IPF meets, removed duplicate 06 SJ/J worlds.


sstangl:
"Open" in meet names is usually part of the official title, used to designate Single-ply meets.

PR #875, opened 2018-03-05
SWales

JoWhiteley:
Results were emailed to me but aren't on the website yet so the URL document is empty


sstangl:
Great! There were some equipment errors -- a number of Raw, Single-ply and Multi-ply lifters toward the end of the document were all marked as Wraps.

Issue #795, opened 2018-02-18
Add FEMEPO (Mexican IPF)


sstangl:
http://www.powerlifting-femepo.com.mx/resultados.html

Their site just shows national results, of which there are only one each year 2009-2017. So not a huge effort to be complete for this federation.

There are state affiliates, but I haven't looked into them. It would be best to get guidance from a Mexican powerlifter.

PR #893, opened 2018-03-08
All of FEMEPO

photon-rain:
Hola!


photon-rain:
Closes #795.

lifter-data/social-instagram.csv
```diff
@@ -382,7 +382,7 @@ Ed Knoblock,ed.k52
 Edan Davey,edandavey
 Eddie Berglund,strongeddie
 Eduard Khanjyan,eduardkhanjyan
-Eduardo Garcia,kimbo_halfslice_
+Eduardo García,kimbo_halfslice_
```
sstangl: I think this is an error -- there's one lifter named "Eduardo Garcia" who competes USPA, and one named "Eduardo García" who competes FEMEPO.

We don't have a way to handle disambiguation when one uses accents and the other doesn't. For the moment, it may be best to keep them as Eduardo Garcia.

The other name corrections are probably multiple lifters under one name also.

PR #899, opened 2018-03-09
Add the data page with same content as prd site. 

MichaelVessia:
TODO: file sizes for downloads


sstangl:
Thank you, everything looks good! :+1:

PR #904, opened 2018-03-09
880: Add check that meet date is not in the future.

fishd:
880: Add check that meet date is not in the future.

Simple datetime comparison to make sure that the meet date isn't more than two days in the future. "The future" is defined as two days to keep timezone management simple.


sstangl:
Thanks!

Issue #908, opened 2018-03-09
Add a test for date and federation name being in the MeetName


sstangl:
The MeetName in `meet.csv` should not contain year or federation information.

For example, some MeetNames look like: "2015 SPF March Madness". Instead, the MeetName should just be "March Madness", with the rest of the information coming from the other fields.

The beta server tries to combine all the fields helpfully, but it would name the above meet the "2015 SPF 2015 SPF March Madness."

MichaelVessia:
Possibly helpful regex, using SPF as an example:
`\d{4}|\b(?:SPF)\b`

PR #913, opened 2018-03-10
Feature/test date fed name

MichaelVessia:
#908

### DIFF
--- a/tests/check-meet-csv
+++ b/tests/check-meet-csv
@@ -369,7 +369,6 @@ def check(scriptname, filename):
             pwarning("The meet %s contains either the year or the federation name" % row[nameidx])
 
 
-
 if __name__ == '__main__':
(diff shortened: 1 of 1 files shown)


sstangl:
Only if it were capitalized, right? It's looking for the string "APA", which doesn't match "Apa".

photon-rain:
True - comparing the lower case version is probably a better idea though, since having the lowercase fed in the MeetName is not unthinkable.

MichaelVessia:
If I use the regex listed in the issue would these problems be avoided?

sstangl:
I think the current version is OK. I've never seen anyone use the lowercase abbreviation.

sstangl:
No -- the regex would cause the problem @photon-rain was trying to avoid above.

sstangl:
Thanks!

Issue #884, opened 2018-03-06
Track WPAU World Powerlifting Alliance Ukraine


TrystanOakley:
Lifter @pechkin.p on instagram requested that we start tracking the WPAU federation. I believe this is their site http://wpa-ukraine.com/ but the only sites he linked me to were all powerlifting. He told me that the president of the Macaw in Ukraine would possibly like to help. His name is Kotsaga Vadim and his instagram is @vadymkotsaga

sstangl:
Results are here: http://wpa-ukraine.com/rezultaty/

TrystanOakley:
@sstangl oh thanks for adding that. I was coming here now to add it myself but you beat me to it lol.

TrystanOakley:
Should I message Vady from that fed and see if he would like to help his fed or do you want to?

sstangl:
No, I don't think there's much use in asking for help at the moment.

PR #928, opened 2018-03-13
WPAU 2017

photon-rain:
All WPAU meets for 2017 +probe & import script. Had to modify calc-latin names to map 'є'->'e' and ''і'->'i'.


photon-rain:
#884.

sstangl:
The "Tested" column is missing, but that's OK since we can derive it from the Pro/Amateur labeling.

PR #938, opened 2018-03-14
Update name-corrections.dat

TrystanOakley:
James Brueggeman prefers to go by Jay. Both names are in the database. This will combine them. @sstangl


sstangl:
@TrystanOakley Wait, that doesn't seem right. You changed him to be known as "Jacob Brueggeman." In the first comment you said his name is "James", but that he prefers to go by "Jay".

Is his name James or Jacob?

PR #945, opened 2018-03-15
WPAU 2015

photon-rain:


sstangl:
I didn't review these too carefully, but the things I skimmed through looked correct. Thanks!

PR #947, opened 2018-03-16
Another ~200 fixes

photon-rain:


meet-data/uspa/0715/meet.csv
```diff
@@ -1,2 +1,2 @@
 Federation,Date,MeetCountry,MeetState,MeetTown,MeetName
-USPA,2016-10-29,USA,UT,Ogden,Halloween Classic
\ No newline at end of file
+USPA,2017-03-18,USA,MO,O'Fallon,3rd Powerbody USA Open
```
sstangl: This is incorrect. USPA uses their own meet numbering system, which we copy -- http://uspa.net/competition_results.html

Meet number `0715` in their database is indeed the Halloween Classic.

I think in the future, for federations for which we just use the normal YYxx numbering scheme, it's OK to have holes and we shouldn't move meets to fill them in.

meet-data/uspa/0799/meet.csv
```diff
@@ -1,2 +0,0 @@
-Federation,Date,MeetCountry,MeetState,MeetTown,MeetName
-USPA,2017-03-18,USA,MO,O'Fallon,3rd Powerbody USA Open
```
sstangl: The existing meet file was correctly numbered as per http://uspa.net/competition_results.html.

PR #965, opened 2018-03-20
Add IG for Ramon Orlanes and Nazar Hrynkiv

MichaelVessia:


MichaelVessia:
RPS made a mistake on Nazar's name will re-open with that fix too

MichaelVessia:
Actually, I don't know if it's a mistake.  I'll fix later if it is

PR #997, opened 2018-03-22
Epf 2002

photon-rain:
Added 2002 EPF Championships.


sstangl:
Thanks! Out of curiosity, why 2002? Was there some standout performance?

photon-rain:
Andrzej Stanaszek's squat record

Issue #887, opened 2018-03-07
Add federation config functionality 


photon-rain:
It would be useful to have a per federation configuration file for the tests. This would contain parameters such as weight classes, divisions and events contested.

sstangl:
Since it's possible that we may port the testing infrastructure to Rust, JSON would be a good choice.

photon-rain:
The ability to exclude meets from certain tests would also be useful - for example 1989 World games has a men's 90+ class, which triggers a sex- weight-class warning.

photon-rain:
This would also be a good first step towards standardisation of division names for each federation.

PR #1012, opened 2018-03-26
Federation config

photon-rain:
Allows federations to be configured. Standardised divisions & weight classes for the IPF and NZPF.


photon-rain:
Not 100% Happy with the formatting of the config files and code so would appreciate your feedback @sstangl.

photon-rain:
Closes #887.

sstangl:
How about https://github.com/toml-lang/toml?

That would look something like:

```
[divisions]
open = {name = "Open", min = 0, max = 999}
juniors = {name = "Juniors", min = 20, max = 23}
masters1 = ["Masters 1", 40, 49] # Alternate syntax, maybe better?

[weightclasses]
# I don't think divisions should be repeated here -- too annoying to maintain.

[exclusions]
1212 = ["check_row_weightclass_sex"] # TOML allows this syntax.
```

photon-rain:
I'll take a look at the TOML format. The reason I was including divisions in weight-classes is to deal with things like the 53kg/43kg IPF classes, which are only options for juniors - this isn't really an essential feature though, so we can probably go without.

photon-rain:
@sstangl How does this look? [CONFIG.txt](https://github.com/sstangl/openpowerlifting/files/1862905/CONFIG.txt)

tests/check-entries-csv
```diff
@@ -681,35 +784,70 @@ def check_row(row, csv):
...
             if 'entries.csv' in files:
                 entriespath = dirname + os.sep + 'entries.csv'
-                check(sys.argv[0], entriespath)
+
+
+                meetpath =''
+                if os.path.isfile(os.path.dirname(dirname)+os.sep+'CONFIG.toml'):
+                    if configpath != os.path.dirname(dirname)+os.sep+'CONFIG.toml': #Only load the file if it has changed
+                        configpath =os.path.dirname(dirname)+os.sep+'CONFIG.toml'
+                        configfile = open(configpath, 'r',encoding='utf-8')
```
sstangl: I think this may leak file handles (since there is no close() if `configfile != None`).

tests/check-entries-csv
```diff
@@ -681,35 +784,70 @@ def check_row(row, csv):
...
+        configtoml = None
+        configfile = None
         for dirname, subdirs, files in os.walk(os.getcwd()):
             if 'entries.csv' in files:
                 entriespath = dirname + os.sep + 'entries.csv'
-                check(sys.argv[0], entriespath)
+
+
+                meetpath =''
+                if os.path.isfile(os.path.dirname(dirname)+os.sep+'CONFIG.toml'):
```
sstangl: I'm a little confused how this works... this is checking for the existence of a file named `CONFIG.toml` in the same directory as the `entries.csv` file. Would it not be in the parent directory? We should probably walk up at least a few parents...

But this is doing a bunch of checks for each entries.csv. Instead, since the tree is visited in depth-first order, couldn't we just check for the presence of `CONFIG.toml` in `files`, and if so, update it statefully and remember that? Then it would be true for all visited children. If we walk to a directory that is outside of that, it would have to be forgotten, but that's easy to check without making a syscall.

meet-data/nzpf/1214/entries.csv
```diff
@@ -9,4 +9,4 @@ Tania Pitt-Guthrie,Masters 1,F,63,100,70,125,295,Raw,SBD,1
 Amanda Young,Masters 1,F,84,105,77.5,140,322.5,Single-ply,SBD,1
 Bruce Park,Masters 5,M,74,115,70,170,355,Raw,SBD,1
 Steve Lousich,Masters 1,M,120+,300,200,230,730,Raw,SBD,1
-Steve Varga,Masters 1,M,110,300,210,260,770,Single-ply,SBD,1
+Steve Varga,Masters 1,M,120,300,210,260,770,Single-ply,SBD,1
```
sstangl: That seems like a big change! Why is that correct?

meet-data/nzpf/1220/entries.csv
```diff
@@ -2,16 +2,16 @@ Name,Division,WeightClassKg,BestSquatKg,BestBenchKg,BestDeadliftKg,TotalKg,Place
...
 Stephen Pickens,Masters 2,74,190,122.5,200,512.5,3,Single-ply,SBD,M
 Jamie King,Juniors,74,170,95,245,510,4,Single-ply,SBD,M
 Harry de Lacey,Juniors,74,185,102.5,212.5,500,5,Single-ply,SBD,M
 Rajah Singh,Open,74,140,102.5,185,427.5,6,Single-ply,SBD,M
-Ming Ming Edgar,Blind Athlete (Juniors),74,140,100,160,400,7,Single-ply,SBD,M
-Joshua Cooke,Special Olympic Athlete (Sub-Juniors),74,100,70,115,285,8,Single-ply,SBD,M
-Bradley Caffell,Special Olympic Athlete (Sub-Juniors),74,82.5,45,87.5,215,9,Single-ply,SBD,M
+Ming Ming Edgar,Blind,74,140,100,160,400,7,Single-ply,SBD,M
+Joshua Cooke,Special Olympics,74,100,70,115,285,8,Single-ply,SBD,M
+Bradley Caffell,Special Olympics,74,82.5,45,87.5,215,9,Single-ply,SBD,M
```
sstangl: This change is actually losing information, though... before, we could maybe include them in the Sub-Juniors and Juniors rankings, but now we'd have no idea.

meet-data/nzpf/1319/entries.csv
```diff
@@ -4,7 +4,7 @@ Karen Froggatt,Raw,F,Masters 2,53.32,57,51,57.5,60,62.5,62.5,62.5,1,B
 Kane McKenzie,Raw,M,Juniors,64.28,66,19,92.5,95,-100,95,95,1,B
 Queenie Lawrence,Raw,F,Open,64.46,72,33,55,60,65,65,65,1,B
 Brian Froggatt,Raw,M,Masters 2,73.77,74,55,80,85,90,90,90,1,B
-Wahdi Moradi,Single-ply,M,Juniors,73.92,74,43,130,137.5,-142.5,137.5,137.5,1,B
+Wahdi Moradi,Single-ply,M,Open,73.92,74,43,130,137.5,-142.5,137.5,137.5,1,B
```
sstangl: That doesn't look valid...

meet-data/nzpf/1507/entries.csv
```diff
@@ -10,7 +10,7 @@ Alex Orwin,1992,M,81.9,83,Juniors,130,150,162.5,162.5,100,115,,115,245,-250,-250
 Leethan Tau,1993,M,118.7,120,Juniors,210,-220,220,220,130,140,145,145,230,-240,240,240,605,SBD,Raw,2
 Lucas Mansell,1993,M,136.6,120+,Juniors,220,237.5,250,250,130,137.5,145,145,240,261,272.5,272.5,667.5,SBD,Raw,1
 Simon Kelsall,1970,M,92.1,93,Masters 1,170,177.5,182.5,182.5,112.5,117.5,-125.5,117.5,195,-205,205,205,505,SBD,Raw,1
-Rachel Duggan,1972,F,95.2,84+,Masters 3,110,120,125,125,60,66,70,70,120,132.5,140,140,335,SBD,Raw,1
-Les Laban,1969,M,100.4,105,Masters 5,-220,-220,-220,,115,120,,120,235,252.5,-255,252.5,,SBD,Raw,DQ
+Rachel Duggan,1972,F,95.2,84+,Masters 1,110,120,125,125,60,66,70,70,120,132.5,140,140,335,SBD,Raw,1
+Les Laban,1969,M,100.4,105,Masters 1,-220,-220,-220,,115,120,,120,235,252.5,-255,252.5,,SBD,Raw,DQ
```
sstangl: This one looks like a good catch, although it could also be a typo in the BirthYear.

meet-data/nzpf/1214/entries.csv
photon-rain: The 110 class was removed and Steve Varga was competing at ~110kg for 2010-2008. He was also 120 for every meet after this.

meet-data/nzpf/1220/entries.csv
photon-rain: Good point! Will add Sub-Junior, Junior subdivisions for Special Olympics/Blind.

meet-data/nzpf/1319/entries.csv
photon-rain: His age is down as 43, so we could go either way on this one.

meet-data/nzpf/1507/entries.csv
photon-rain: From the results file, this was a mistake when I was adding the meet.

tests/check-entries-csv
photon-rain: It does indeed, will fix.

photon-rain:
I'm not really sure how a mapping layer would work, you'd have to account for all the possible variations of a division name. For most meets we should have a results file, so I don't think losing track of changes should be too huge an issue.

sstangl:
OK, that sounds sane. Woohoo! This should help quite a lot with maintaining some sanity. Thank you so much for this work.

PR #1042, opened 2018-03-30
2017 Polish Raw Nationals

photon-rain:
Has Krzysztof Wierzbicki's WR total.


sstangl:
The federation appears to have named themselves `PZKFiTS` (with a lowercase `i`), so I think we should keep that. I'll change. Thanks for adding :+1:

PR #1045, opened 2018-03-31
WRPF Corrections

photon-rain:
I think this fixes all the Raw/Wraps errors.


sstangl:
Thank you, and sorry about the hassle! Do the single-ply results still need correction, or are we done with WRPF?

photon-rain:
The single-ply results should also be fixed - only found the one error.

PR #1046, opened 2018-03-31
FPR Nationals 2007-2018

photon-rain:
Missing 2013 since it was a scan of a PDF.


photon-rain:
This'll probably be me done with Russia for a while - hopefully what we've got up entices someone else to help out.

sstangl:
I don't expect any Russians to help out until we have the site fully translated into Russian, but this is indeed a sufficient mass of results. Thank you!

PR #1055, opened 2018-04-03
Update name-corrections.dat

emp100k:


sstangl:
Some of these changes should be made in the original files (in the cases of the reversed names).

PR #1061, opened 2018-04-04
Update entries.csv

emp100k:
These three entries had no total and a huge deadlift number due to a comma spacing error placing deadlifts into the bench press column and totals into the deadlift column.  I do not know of these lifters were actually disqualified as they did form complete totals, there doesn't appear to be any DQ indicator on the results sheet but I will leave that as the placing for know until I can confirm.  If DQ was added during formatting as an result of there being no number in the total column then I will infer that these 3 lifters were in fact not disqualified.


sstangl:
Yeah, the DQ was placed automatically. So they probably weren't disqualified. Good catch!

emp100k:
Looks like the build check won't allow it with DQ on a row that has a total...  
Also of note there was another kid that recorded a lift in all 3 disciplines in this meet but had no total shown in the results file so I left him as is...

emp100k:
Do I have to re-submit this or did you handle it?

sstangl:
I'll fix it when merging.

emp100k:
Thanks!
    On Wednesday, April 4, 2018, 2:34:19 PM EDT, Sean Stangl <notifications@github.com> wrote:  
 
 
I'll fix it when merging.

PR #1063, opened 2018-04-05
Powerlifting at the Rio 2016 Paralympic Games

emp100k:
I think I finally figured out how to add meets just with the online github site, this is for the ParaPL meet (bench only) at the 2016 Rio Paralympics.


photon-rain:
I've combined all the files into one request - could you add the missed attempts to the entries files?

emp100k:
yeah no problem

photon-rain:
Cool, thanks.

emp100k:
For missed attempts we put a - in front, but for untaken attempts we just leave it blank right?

sstangl:
@emp100k Yes

PR #1069, opened 2018-04-06
Standardised PA divisions

photon-rain:


sstangl:
I was hesitant about this change, but upon further review it seems safe and will let the Rust code use an enum.

In general I prefer keeping to whatever names the federation has -- so if they use "M-RO" or something like that, I prefer to keep that instead of just "Open". But in this case it's roughly equivalent.

photon-rain:
PA seems to be really inconsistent in the way that they style their divisions, so I just settled on using the IPF divisions. For federations like USAPL we should probably standardise on the "M-RO" format.

PR #1073, opened 2018-04-09
Added a few meets, no text files(sorry) and needs to be converted to kgs(sorry!)

xxxdolorxxx:
Hey guys, I added a few IPA meets to the best of my abilities. 
Unfortunately there's going to be some issues. 
1. There's no text files. I'm sorry, I'm still learning. 
2. The meets are going to have to be converted to kgs. Again my apologies for being a noob with the coding. 

Thank you!


sstangl:
Thanks!

While reviewing, I noticed that there were a significant number of typos in the data. Some lifter bodyweights were incorrect, individual lifts had numbers swapped around, and names were misspelled.

This is not a problem with you -- you actually have a better-than-expected error rate. The problem is that _any_ attempt to enter in data manually, for example by retyping a PDF, will introduce a large number of errors. We unfortunately can't permit _any_ errors. Therefore we don't enter data in manually.

The trick we use is to download the data from the federations and then do transforms on that data, using whatever tools get the job done, to convert it into our format. For the case of the IPA, I've had success with using the tool `pdftotext` with the `-layout` option, and using a text editor after that.

When done this way, none of the numbers are re-typed, and none of the names are re-typed, so the error rate tends to approximate zero.

These commits are close to correct, though! I'll fix them up and merge them in.

If you would like to join our team chat, where it's easier to talk about these things, please mail updates@openpowerlifting.org.

PR #1075, opened 2018-04-09
Rio 2016

emp100k:
Added missed attempts.

### DIFF
--- a/lifter-data/name-corrections.dat
+++ b/lifter-data/name-corrections.dat
@@ -106,6 +106,7 @@ Alexey Sivokon,Alexei Sivokon,Aleyey Sivokon
 Alexis Guzmán,Alexis Guzman
 Alf-Ivar Hellerud,Alf Ivar Hellerud
 Ali Al Lami,Ali Al-Lami
+Ali Sadeghzadeh Salmani,Ali Sadeghzadehsalmani
 Alicia Troxtell,Alicia Traxtell
 Alice MacKenzie,Alice Mackenzie
 Aline Gonçalves,Aline Goncalves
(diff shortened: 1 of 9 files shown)


emp100k:
looks like it doesn't like the 4th attempts...  total = best of first 3 attempts but bestbenchkg needs to be the highest bench including 4th...

photon-rain:
BestBenchKg is the same as the total in this case, we really should rename it Bestof3Kg or something like that.

emp100k:
Right but then won't the 4th attempts not show up on the ranking list?

photon-rain:
I think they still show up - @sstangl how does this work?

emp100k:
the only other error I see for this and the meets.csv is requiring the other file to already exist, but doesn't one need to be uploaded first?

photon-rain:
I'll combine them into one pull request - I've just checked and 4th attempts do show up on the site.

emp100k:
So then I just need to change the bestbenchkg column to equal the totalkg column?

sstangl:
@emp100k @photon-rain The site takes the best of the `BestSquatKg` and `Squat4Kg` columns when displaying the column "Squat".

`BestSquatKg` should be renamed to `Best3SquatKg` -- I'll do this soon. It's the best of the attempts that counted toward the `TotalKg`.

emp100k:
So then in this case I should just make the bestbenchkg of those lifters that made 4th attempts be equal to their totalkg right?

sstangl:
If it's bench-only, yeah, `BestBenchKg = TotalKg`.

photon-rain:
You can go down in weight for 4th attempts in Para-PL? That's bizarre.

emp100k:
Yeah i noticed that too, guess the standard 3 attempts are rising bar like in any other powerlifting meets but 4th attempts are totally un-related...

PR #1078, opened 2018-04-10
AfricanPF 2006,2007,2009

photon-rain:
There were also championships in 2010 & 2011 but I think they're lost to time.


sstangl:
Awesome, thank you.

PR #1089, opened 2018-04-12
Epa update 2015 part 2

JoWhiteley:
This completes 2015 for EPA


photon-rain:
I forgot to comment this last night - 1525 needs the attempt misses added.

JoWhiteley:
Oh my word.  They aren't marked on the original either.  Thankfully there are MaxFoo columns so should be able to piece it together.

JoWhiteley:
@photon-rain fixed 1525

PR #1099, opened 2018-04-16
2009 part 1

JoWhiteley:


JoWhiteley:
Fixed most of the errors (yay me!) but not sure what to do about the woman whose bodyweight does not match weight class.  This is what the source data says so there's no way of checking.

photon-rain:
Wow this was quick!

PR #1109, opened 2018-04-17
Update entries.csv

emp100k:
a 780lb bench and 2635lb total are a bit out of the scope of what someone can do raw right now... also Troy here has never competed in anything but Multi-ply.


sstangl:
Good ol' SPF.

PR #1110, opened 2018-04-17
Update results.csv

emp100k:
The WeightClassKg for Jonathan Clark and Greg Monmaney are incorrect, they were the only two in this meet that did not have a BodyweightKg and as such the meet's scoresheet software may have defaulted them to the -114 class (smallest male class in the RPS).
Jonathan Clark is typically around 242, and Greg Monmaney is typically 181-198, that said without actually knowing what they weighed in as I can't put anything in the WeightClassKg or BodyweightKg to replace the bad value.  For now just leave WeightClassKg blank?


emp100k:
oops, I changed the results file, not the entries file, I'll make the change in the right file...

PR #1129, opened 2018-04-26
SVNL

photon-rain:
Changed the weightclass column check in check-entries-csv to a row check so that we can don't have to add weight-classes excluded in a config file.


meet-data/nsf/Barents_Games_2016/meet.csv
```diff
@@ -0,0 +1,2 @@
+Federation,Date,MeetCountry,MeetState,MeetTown,MeetName
+SVNL,2017-09-02,Norway,,Bodø,Barents Summer Games
```
sstangl: This should go into the `svnl/` folder based on URL.

meet-data/svnl/03012015-penkkipunnerrus-p%C3%A4rnu-hpv/URL
```diff
@@ -0,0 +1 @@
+https://suomenvoimanostoliitto.fi/tulokset/kansalliset-kilpailut/03012015-penkkipunnerrus-p%C3%A4rnu-hpv
```
sstangl: This `MeetPath` is `svnl/03012015-penkkipunnerrus-p%C3%A4rnu-hpv` which requires double-escaping the `%%` signs, which is pretty ugly.

Also, the dates occur here in a non-standard order to the rest of the site. Since the `MeetPath` is the public name of the meet, that looks pretty inconsistent.

One option is to use the hash of the name, as in `cpu/cpu-reify-database` line 18. Python`s hash function is defined to be stable, and that produces pretty URLs.

A second option is to just use our default `1701`, `1702`, ..., which produces very short URLs.

web/templates/fedselect.frag
```diff
@@ -75,6 +75,7 @@
   </optgroup>
   <optgroup label="Finland">
     <option value="FPO">FPO - Finland Powerlifting Organization [IPA] (Incomplete)</option>
+    <option value="SVNL">SVNL - Suomen Voimanostoliitto [IPF]</option>
```
sstangl: I'm not sure, but I think it should be `Suomen Voimanostoliitto ry` based on their usage.

They should also be added to the list of "drug-tested" federations at the top of this file, and there should also be an "All Finland" option added.

tests/check-entries-csv
```diff
@@ -261,46 +261,46 @@ def check_column_birthyear(s, row, csv):
...
-
-    classes = ipf_men + ipf_women + ipl + rum_extra + proraw_extra + \
-        rps_extra + spf_extra + ipf_extra + ipf_women_extra + \
-        old_usapl + ffforce_extra + wnpf_extra + apa_extra + \
-        idfpf_extra + parapl_extra + wpau_extra + parapl_men + parapl_women
-
-    if s not in classes:
-        pwarning("Unknown weight class: %s" % s)
+def check_row_weightclass(row, csv, weightclasses):
+    if 'WeightClassKg' in csv.fieldnames and row[csv.index('WeightClassKg')] != "":
```
sstangl: Reduce the level of indentation by changing this to a guard:

```
if 'WeightClassKg' not in csv.fieldnames or row[csv.index('WeightClassKg')] == '':
    return
```

Guards are always preferred over indentation.

server/src/opldb/fields/federation.rs
```diff
@@ -154,6 +155,7 @@ impl fmt::Display for Federation {
             Federation::SCT => write!(f, "SCT"),
             Federation::SPA => write!(f, "SPA"),
             Federation::SPF => write!(f, "SPF"),
+            Federation::SVNL => write!(f, "SVNL"),
```
sstangl: Thanks for adding this!

meet-data/svnl/ignore/URL
```diff
@@ -0,0 +1,16 @@
...
+https://suomenvoimanostoliitto.fi/tulokset/sm-kilpailut/1981-voimanoston-sm-kisat-naiset-helsinki
+https://suomenvoimanostoliitto.fi/tulokset/sm-kilpailut/1991998-penkkipunnerruksen-joukkue-sm-loimaa
+https://suomenvoimanostoliitto.fi/tulokset/kansalliset-kilpailut/16052015-klassinen-voimanosto-tammisaari-kk-eken-0
+https://suomenvoimanostoliitto.fi/tulokset/kansalliset-kilpailut/17052014-penkkipunnerrus-klassinen-kitee-papu-0
+https://suomenvoimanostoliitto.fi/tulokset/kansalliset-kilpailut/28122015-klassinen-penkkipunnerrus-pietarsaari-jp-2000-0
+https://suomenvoimanostoliitto.fi/tulokset/sm-kilpailut/2-331985-voimanosto-miehet-tampere
+
+
+
+
```
sstangl: Not that it really matters much, but extraneous whitespace here.

meet-data/svnl/03012015-penkkipunnerrus-p%C3%A4rnu-hpv/URL
sstangl: Thinking about it, I'd probably prefer `1701`, `1702`, etc.

PR #1136, opened 2018-04-30
Sweden

photon-rain:
Server still works too well :wink:


tests/check-meet-csv
```diff
@@ -434,7 +435,7 @@ def check(scriptname, filename):
         # Check if MeetName column contains the year or the fed name
         if year in row[nameidx]:
             perror("MeetName should not have the year.")
-        if row[fedidx] in row[nameidx]:
+        if row[fedidx] in row[nameidx].split():
```
sstangl: Sometimes MeetName is like "USPA/IPL FooBar", so I was hoping to catch that case also.

server/src/opldb/fields/federation.rs
```diff
@@ -142,6 +142,8 @@ pub enum Federation {
     #[strum(to_string = "SPF", serialize = "spf")]
     SPF,
     #[strum(to_string = "SVNL", serialize = "svnl")]
+    SSF,
+    #[strum(to_string = "SSF", serialize = "ssf")]
```
sstangl: The `#[strum]` decorator should go above the line being described. (SVNL needs to be moved back down.)

meet-data/ssf/ssf-probe
```diff
@@ -0,0 +1,103 @@
+#!/usr/bin/env python3
+# vim: set ts=8 sts=4 et sw=4 tw=99:
+#
+# Probes for new meets from the NSF.
```
sstangl: SSF :-)

tests/check-meet-csv
photon-rain: I switched it to this because SSF has a bunch of meets with VSSF, OSSF etc in the names (Which I think are regional affiliates?) I'll add a few replaces to this line to include the cases that you're trying to catch.

sstangl: The server is intended to render meet names as `${year} ${fed} ${meetname}`, like `2018 RPS Spring Supremacy`, which is customary usage.

If a meet includes VSSF, that will render as `2018 SSF VSSF Foo Competition`.

It's your call whether or not that's appropriate.

sstangl: For BP, we do allow regional affiliates like that. So you can get `2018 BP WM Meet`, where "WM" is for the regional affiliate "West Midlands".

photon-rain: From memory the regional identifiers are all at the end of the meet name, so it shouldn't look too strange. There are several meets missing a city, so I think keeping the regional identifiers when they're present will be useful.

photon-rain:
@sstangl Made requested changes.

PR #1139, opened 2018-05-01
Nordic Powerlifting Federation.

photon-rain:
One more regional IPF affiliate to go.


photon-rain:
@sstangl Switch to NordicPF and rebased.

sstangl:
It looks like there may be an error with the rebase -- this commit includes way too many files, like the whole SSF changeset.

PR #1199, opened 2018-05-12
CSST

### LINKED ISSUE #1093: Import Czech IPF results
The Czech IPF affiliate has nicely formatted data which can probably be auto-imported:

Pre 2018:
http://powerlifting.cuscz.cz/vysledky/souteze.htm

2018:
http://www.powerlifting-csst.cz/cz/vysledky-2018.html

Their abbreviation seems to be CSST.

photon-rain:
Finishes all the federations that we know of that use html for results. Closes #1093.


photon-rain:
I didn't add the name corrections in alphabetical order since there were ~600 corrections. We should probably sort name-corrections.dat after this.

sstangl:
I honestly never made any attempt to store that file in sorted order -- it's OK if it's in random order. Probably would be a pain to maintain it, and there wouldn't be any benefit.

photon-rain:
I find it useful for finding lifters with shortened names and middle names.

PR #1206, opened 2018-05-15
epa update

JoWhiteley:


JoWhiteley:
I give up.  Don;t know what;s up with this one.

sstangl:
A lifter's name was entered as `Jack I<92>Anson`, where the `<92>' is a single invalid character using some old-style character encoding.

I believe this should pass tests, although `epa/1816` doesn't have `URL` or `results.txt` files.

PR #1229, opened 2018-05-17
Update birthdays.csv

TrystanOakley:
Added some teammates @sstangl


sstangl:
@TrystanOakley Adding birthdays to the public record shouldn't just be a casual thing we do. If they gave approval for this, then it's fine, but they need to know what they're doing.

TrystanOakley:
Do I need to change something on this pull request or delete it or?

sstangl:
Yeah, I think let's delete it. We'll need to find a better mechanism than `birthdays.csv`.

PR #1239, opened 2018-05-18
Bodypower

JoWhiteley:
BPU British Championships


JoWhiteley:
querying some results.  leave this hanging for the moment.  I will sort it tomorrow

JoWhiteley:
@sstangl if this failed due to name corrections - is that something I can fix from either the desktop client or the online client?

PR #1242, opened 2018-05-19
WPO meets

photon-rain:
All WPO meets apart from the 2001/2000 meets which I can't find results for. WPO merged with the WPC at some point- so I've logged these as WPC, but left WPO in the name.

Also included Vlad Alhazov's 1250lb squat and Andy Bolton's 1008lb Deadlift.

This breaks our Wilks formula implementation, will be fixed in the following pull.


sstangl:
OK, that should be fine. We might at a future point add WPO as a federation and just make WPC a joint federation, like was done with ADFPA/USAPL.

PR #1256, opened 2018-05-22
SE 2013

JoWhiteley:


JoWhiteley:
@sstangl OK fixed errors.  There's a problem with epa/1312 though.  Each lifter had 4 attempts and the BestBench and Total column reflect the highest of the 4.  This is correct as it is the way the competition was run.  However it is failing the checks.  Not sure of the way forward here.

Issue #939, opened 2018-03-14
Lifter Request: Add meet for Richard Hawthorne


TrystanOakley:
A meet from 2013 is missing for Richard Hawthorne. The meet took place on August 10th and 11th of 2013 at the CAPO National Powerlifting Championships located at the Derwent Ent. Center in Hobart, Tasmania.

The link http://globalpowerliftingalliance.com/results.html is reported as being "dead" for this meet so the way back machine may have to be used to get the results.

photon-rain:
This meet is 13 jpgs :/ https://web.archive.org/web/20160404190038/http://capoaustralianpowerlifting.com.au/index.php/calendar-2/results

sstangl:
Boo, Australia. Boo.

PR #1305, opened 2018-06-02
All time records.

### LINKED ISSUE #1079: Find and add 2014 Dubai Fazza games results (parapl)
ok I have a challenge for you guys, can you find this meets results: https://www.paralympic.org/news-1824  It's name was the 4th Annual Fazza International games in Daubai in Feb 2012.  I have been searching for this meet report for the better part of a year to no avail.  This article I linked lists the results of the top 3 in each weight class but that's not a complete results...  :disappointed:    It's pretty relevant because this meet is where the single lift bench WR was set for woman's -148 and men's -123.

photon-rain:
Added most of the records from the PLW lists. Closes #1079, #939, finishes one of the meets in #849.


photon-rain:
I needed to configure APA for this, so check that you agree with my division labeling.

sstangl:
`server/templates/widget-fedselect.tera` also needs an update. Sorry that we have to maintain it in two places!

photon-rain:
Ah, thought there was a third template file, just couldn't find it 😄. I'll switch the spelling to submasters if that's your preferred nomenclature. Will write an APA division script next week, this is something that we'll need to have for most federations so a shared codebase will probably be useful (like what we have with oplprobe).

PR #1314, opened 2018-06-03
Rename column `BestFooKg` to `Best3FooKg`.

sstangl:
The previous naming consistently caused confusion when using our data.
People were not sure whether 4th and 5th attempts should be counted
in the "Best" column. Renaming to "Best3" should make it more obvious
that this is intended to just be the best-of-the-first-three attempts.

We've had internal code errors also where I've gotten confused myself.
Hopefully this is the end of that!


sstangl:
@JoWhiteley This change is likely to trip us all up for a bit.

PR #1319, opened 2018-06-06
Update name-corrections.dat

jwakefieldau:
Fixing some incorrect names noticed in GPC Aus 2018 Nats, also, Lauren Hannon recently married and is now Lauren Green


sstangl:
Resolved on GitLab.

